PR #1, opened 2017-09-27
Slight refactor, remove Redshift-specific items from main code

jameremo:
This PR is part 1 of 2 parts to the effort to make hauser support BigQuery.  It's just a slight refactor of stuff that is either directly or indirectly Redshift-specific.  After we're good with this, I'll come along with another PR that actually has the BQ code (which I'm still testing locally).

### DIFF
--- a/main.go
+++ b/main.go
@@ -9,7 +9,6 @@ import (
 	"log"
 	"os"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"./config"
@@ -26,29 +25,11 @@ var (
 // Represents a single export row in the export file
 type Record map[string]interface{}
 
-func ValueToString(wh warehouse.Warehouse, val interface{}, f warehouse.Field) string {
-	s := fmt.Sprintf("%v", val)
-	if f.IsTime() {
-		t, _ := time.Parse(time.RFC3339Nano, s)
-		return t.String()
-	}
-
-	s = strings.Replace(s, "\n", " ", -1)
-	s = strings.Replace(s, "\x00", "", -1)
-
-	if maxlen, ok := wh.VarCharMaxLen(); ok == true {
-		if len(s) >= maxlen {
-			s = s[:maxlen-1]
-		}
-	}
-	return s
-}
-
 func TransformExportJsonRecord(wh warehouse.Warehouse, rec map[string]interface{}) ([]string, error) {
 	var line []string
 	for _, field := range warehouse.ExportTableSchema {
 		if val, ok := rec[field.Name()]; ok {
-			line = append(line, ValueToString(wh, val, field))
+			line = append(line, wh.ValueToString(val, field))
 			delete(rec, field.Name())
 		} else {
 			line = append(line, "")
--- a/main.go
+++ b/main.go
@@ -27,15 +27,20 @@ type Record map[string]interface{}
 
 func TransformExportJsonRecord(wh warehouse.Warehouse, rec map[string]interface{}) ([]string, error) {
 	var line []string
-	for _, field := range warehouse.ExportTableSchema {
-		if val, ok := rec[field.Name()]; ok {
+	for _, field := range wh.ExportTableSchema() {
+		if field.IsCustomVars {
+			continue
+		}
+
+		if val, ok := rec[field.Name]; ok {
 			line = append(line, wh.ValueToString(val, field))
-			delete(rec, field.Name())
+			delete(rec, field.Name)
 		} else {
 			line = append(line, "")
 		}
 	}
 
+	// custom variables will be whatever is left after all well-known fields are accounted for
 	customVars, err := json.Marshal(rec)
 	if err != nil {
 		return nil, err
--- a/main.go
+++ b/main.go
@@ -26,9 +26,9 @@ var (
 // Represents a single export row in the export file
 type Record map[string]interface{}
 
-func ValueToString(val interface{}, f warehouse.Field) string {
+func ValueToString(wh warehouse.Warehouse, val interface{}, f warehouse.Field) string {
 	s := fmt.Sprintf("%v", val)
-	if f.DBType == "TIMESTAMP" {
+	if f.IsTime() {
 		t, _ := time.Parse(time.RFC3339Nano, s)
 		return t.String()
 	}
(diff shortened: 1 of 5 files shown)


warehouse/redshift.go
```diff
@@ -246,3 +251,23 @@ func (rs *Redshift) DoesTableExist(name string) bool {
...
 	return (exists != 0)
 }
+
+func toColumnDef(field Field) string {
+	var dbType string
+	switch field.DataType().Kind() {
+	case reflect.Int64:
+		dbType = "BIGINT"
+	case reflect.String:
+		dbType = "varchar(max)"
```
jerseybbq: capitalize `varchar` for consistency

warehouse/schema.go
```diff
@@ -1,53 +1,90 @@
...
-	Name   string
-	DBType string
+func (f *dataField) Name() string {
+	return f.name
 }
 
-func (f Field) String() string {
-	return fmt.Sprintf("%s %s", f.Name, f.DBType)
+func (f *dataField) DataType() reflect.Type {
+	return f.dataType
```
jerseybbq: It might be more clear to have this method return `reflect.TypeOf(time.Now())` when appropriate and get rid of `IsTime()`

warehouse/schema.go
```diff
@@ -1,53 +1,90 @@
...
+}
+
+func (f *dataField) IsTime() bool {
+	return f.dataType.Kind() == reflect.Struct && f.dataType.PkgPath() == "time" && f.dataType.Name() == "Time"
```
jerseybbq: Why wouldn't this be `return (reflect.TypeOf(f.dataType) == reflect.TypeOf(time.Time))`?

warehouse/schema.go
jameremo: I apologize, I had decided that since you have to check for a time data type on every field on every row, I did not want to always repeat that work.  I was refactoring to do the check once when the Field slices were first created, and didn't quite finish.  Now the refactor is complete, but I can see there is an argument to be made for not having a special `IsTime()` function, given that you could end up with a whole bunch of `IsSomeDataType()` functions moving forward.

warehouse/schema.go
jameremo: This is just my relative ignorance of reflection in Go, I've updated the check in the newly added `newDataField()` function.

jameremo:
@mgm PTAL, when you have a free minute

warehouse/schema.go
```diff
@@ -1,53 +1,93 @@
 package warehouse
 
-import "fmt"
+import (
```
jerseybbq: Warning: long-ish comment!

I think we can simplify this a bit and make it more Go like. In the current version, responsibility is spread a bit too much - there's a lot of type mapping here, but then in a specific provider (like `redshift.go`) we're still converting things, looping and mapping to strings, etc. That'll end up being boilerplate that other providers will have to implement as well, so it's nicer to do it once in a single place.

Rough code below is how it looks. Things to point out: the central idea here is that a provider just specifies a Go-to-DB type mapping, eg:

```
var (
	RedshiftTypeMap = FieldTypeMapper{
		"int64":     "BIGINT",
		"string":    "VARCHAR(max)",
		"time.Time": "TIMESTAMP",
	}
)
```
which is passed in to `ExportTableSchema()` to produce a provider-specific schema. (Another advantage of this approach is easy re-use of common type mappers.) 

The Stringers for Field and Schema do the heavy lifting for `create table` type statements - that won't be suitable for every provider, but it's a common case and easy to do here. If something more custom is needed (in BQ it's definitely the case), we can do that in the provider.   

I stripped down the struct types a bit, and removed getters...in Go, it's generally preferable to give direct access to struct fields if it's safe. Also, I put `CustomVars` into `exportSchema` - no need to separate them when formulated this way (it was vestigial from my first version).

There are a couple of ripple changes from this into `main.go`, but it's very minor - we can look at that together.


```
package warehouse

import (
	"fmt"
	"log"
	"reflect"
	"strings"
	"time"
)

type exportSchema struct {
	EventStart             time.Time
	EventType              string
	EventTargetText        string
	EventTargetSelectorTok string
	EventModFrustrated     int64
	EventModDead           int64
	EventModError          int64
	EventModSuspicious     int64
	IndvId                 int64
	PageUrl                string
	PageDuration           int64
	PageActiveDuration     int64
	PageRefererUrl         string
	PageLatLong            string
	PageAgent              string
	PageIp                 string
	PageBrowser            string
	PageDevice             string
	PageOperatingSystem    string
	PageNumInfos           int64
	PageNumWarnings        int64
	PageNumErrors          int64
	SessionId              int64
	PageId                 int64
	UserAppKey             string
	UserEmail              string
	UserDisplayName        string
	UserId                 int64
	CustomVars             string
}

type syncTable struct {
	ID            int64
	Processed     time.Time
	BundleEndTime time.Time
}

type Field struct {
	Name   string
	DBType string
	IsTime bool
}

func (f Field) String() string {
	return fmt.Sprintf("%s %s", f.Name, f.DBType)
}

type Schema []Field

func (s Schema) String() string {
	ss := make([]string, len(s))
	for i, f := range s {
		ss[i] = f.String()
	}
	return strings.Join(ss, ",")
}

type FieldTypeMapper map[string]string

func ExportTableSchema(ftm FieldTypeMapper) Schema {
	return structToSchema(exportSchema{}, ftm)
}

func SyncTableSchema(ftm FieldTypeMapper) Schema {
	return structToSchema(syncTable{}, ftm)
}

func structToSchema(i interface{}, ftm FieldTypeMapper) Schema {
	t := reflect.TypeOf(i)
	result := make(Schema, t.NumField())
	for i := 0; i < t.NumField(); i++ {
		result[i] = Field{
			Name:   t.Field(i).Name,
			DBType: convertType(ftm, t.Field(i).Type),
			IsTime: t.Field(i).Type == reflect.TypeOf(time.Time{}),
		}
	}
	return result
}

func convertType(ftm FieldTypeMapper, t reflect.Type) string {
	dbtype, ok := ftm[t.String()]
	if !ok {
		log.Fatal("Type %s is not present in FieldTypeMapper", t)
	}
	return dbtype
}
```

main.go
```diff
@@ -27,15 +27,20 @@ type Record map[string]interface{}
 
 func TransformExportJsonRecord(wh warehouse.Warehouse, rec map[string]interface{}) ([]string, error) {
 	var line []string
-	for _, field := range warehouse.ExportTableSchema {
-		if val, ok := rec[field.Name()]; ok {
+	for _, field := range wh.ExportTableSchema() {
+		if field.IsCustomVars {
+			continue
+		}
```
jameremo: It turns out that having the CustomVars separate from the main schema wasn't quite vestigial; it was necessary to make the CustomVars JSON generated in this function work.  I wanted to handle the JSON here rather than each warehouse impl, so I added a new field to the Field struct to indicate whether the field is the designated "CustomVars" field.

warehouse/schema.go
```diff
@@ -84,10 +46,55 @@ type syncTable struct {
 	BundleEndTime time.Time
 }
 
-func toFieldSlice(t reflect.Type) []Field {
-	result := make([]Field, t.NumField())
+type Field struct {
+	Name         string
+	DBType       string
+	IsTime       bool
+	IsCustomVars bool
```
jerseybbq: Nit: should be singular (`IsCustomVar`)

PR #2, opened 2017-10-13
Added flag to upload files to S3 without loading to Redshift

jameremo:


warehouse/redshift.go
```diff
@@ -135,22 +135,27 @@ func (rs *Redshift) LoadToWarehouse(file string) error {
...
 
-	rs.conn, err = rs.MakeRedshfitConnection()
-	if err != nil {
-		return err
-	}
-	defer rs.conn.Close()
+	if rs.conf.S3.S3Only {
+		log.Printf("Config flag S3Only is on, skipping load to Redshift")
+	} else {
+		defer rs.RemoveS3Object(s3obj)
```
jameremo: I considered adding a second flag that allows you to turn off file deletion even if you're loading into Redshift.  Not sure if anyone would need that unless they were debugging, though.

jerseybbq: Rather than put the redshift logic in an `else` clause, it would be cleaner to return from the `if rs.conf.S3.S3Only` clause. Also, rather than log `skipping load to redshift` on each file, I'd just do it once on startup (since the flag isn't dynamic).

jerseybbq: agreed

jameremo:
PTAL @jerseybbq, when you have a minute.

warehouse/redshift.go
```diff
@@ -35,6 +35,7 @@ var (
 )
 
 func NewRedshift(c *config.Config) *Redshift {
+	log.Printf("Config flag S3Only is on, data will not be loaded to Redshift")
```
sanjaybv: @jameremo I don't understand how the config here has `S3Only` set to `true` always?

jameremo: It doesn't, this is a completely embarrassing bit hurried coding without testing.  😬 Thank you for catching it.

PR #3, opened 2017-10-13
Optionally group bundles by day for upload

jameremo:
This PR adds a new flag called `GroupFilesByDay` which, when true, will cause Hauser to combine all the bundles for the same day into a single CSV file before loading the file into the warehouse.  Grouping the bundles together provides two benefits, particularly when there are many small bundles:

1.  Loading data for many days at once is much faster, as the number of round trips from Hauser to the warehouse is greatly reduced.
2.  For BigQuery specifically, there is a quota of 1000 load jobs per day.  It is not uncommon for larger clients to generate bundles every 30 minutes, which means 48 bundles per day.  Thus you cannot load more than 20-21 days of data before you reach the limit.  Grouping the bundles allows you to load much more data before using up the quota.

### DIFF
--- a/main.go
+++ b/main.go
@@ -62,8 +62,18 @@ func ProcessExportsSince(wh warehouse.Warehouse, since time.Time) (int, error) {
 		return 0, err
 	}
 
-	for _, e := range exports {
-		log.Printf("Processing bundle %d (start: %s, end: %s)", e.ID, e.Start, e.Stop)
+	if conf.GroupFilesByDay {
+		return LoadFilesGrouped(wh, fs, exports)
+	} else {
+		return LoadFilesSingly(wh, fs, exports)
+	}
+}
+
+// LoadFilesSingly iterates over the list of available export files and processes them one by one, until an error
+// occurs, or until they are all processed.
+func LoadFilesSingly(wh warehouse.Warehouse, fs *fullstory.Client, exports []fullstory.ExportMeta) (int, error) {
+	for i, e := range exports {
+		log.Printf("Processing bundle %d (start: %s, end: %s)", e.ID, e.Start.UTC(), e.Stop.UTC())
 		filename := filepath.Join(conf.TmpDir, fmt.Sprintf("%d.csv", e.ID))
 		mark := time.Now()
 		outfile, err := os.Create(filename)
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -75,57 +85,17 @@ func ProcessExportsSince(wh warehouse.Warehouse, since time.Time) (int, error) {
...
 
-		csvOut.Flush()
-
 		if err := wh.LoadToWarehouse(filename); err != nil {
 			log.Printf("Failed to load file '%s' to warehouse: %s", filename, err)
 			return 0, err
 		}
 
-		if err := wh.SaveSyncPoint(e.ID, e.Stop); err != nil {
+		if err := wh.SaveSyncPoints(exports[i : i+1]); err != nil {
```
jerseybbq: Eww. ;) 

That's tricky to read. I suggest changing `SaveSyncPoint` to a variadic function so it reads cleanly (and you don't even have to change the logic). So like:

`func (rs *Redshift) SaveSyncPoint(bundles ...fullstory.ExportMeta) error `

and then the call here is just: `if err := wh.SaveSyncPoint(e); err != nil {`

and the corresponding call in the grouped version is `wh.SaveSyncPoint(processedBundles...)`

main.go
```diff
@@ -142,6 +112,110 @@ func ProcessExportsSince(wh warehouse.Warehouse, since time.Time) (int, error) {
...
+		log.Printf("Failed gzip reader: %s", err)
+		return 0, err
+	}
+
+	decoder := json.NewDecoder(gzstream)
+	decoder.UseNumber()
+
+	// skip array open delimiter
+	if _, err := decoder.Token(); err != nil {
+		log.Printf("Failed json decode of array token: %s", err)
```
sanjaybv: NIT Maybe mention which array token because the error message below is the same as well.

main.go
jameremo: Thanks, that's wonderfully elegant.

main.go
```diff
@@ -49,7 +49,7 @@ func TransformExportJsonRecord(wh warehouse.Warehouse, rec map[string]interface{
 	return line, nil
 }
 
-func ProcessExportsSince(wh warehouse.Warehouse, since time.Time) (int, error) {
+func ProcessExportsSince(wh warehouse.Warehouse, since time.Time, processFunc func(warehouse.Warehouse, *fullstory.Client, []fullstory.ExportMeta) (int, error)) (int, error) {
```
jerseybbq: Nit: you may want to type alias the function signature to make it easier to reference.

main.go
```diff
@@ -240,6 +236,13 @@ func main() {
 		log.Fatal(err)
 	}
 
+	var processFunc func(warehouse.Warehouse, *fullstory.Client, []fullstory.ExportMeta) (int, error)
```
jerseybbq: Nit: this could be expressed more cleanly as:
```
processFunc := ProcessFilesIndividually
if conf.GroupFilesByDay {
    processFunc = ProcessFilesByDay
}
```
since processing individually is the default

PR #4, opened 2017-10-16
Add BigQuery warehouse implementation

jameremo:
After much ado and refactoring, support for BigQuery is finally here!  also includes an update to README.md with information relevant to the new implementation.


jameremo:
I rebased this today to incorporate changes from #3, so my apologies if there were comments pending and this messes up lines numbers and such.

warehouse/bigquery.go
```diff
@@ -0,0 +1,394 @@
...
+func (bq *BigQuery) createExportTable() error {
+	log.Printf("Creating table %s", bq.conf.BigQuery.ExportTable)
+
+	schema, err := bigquery.InferSchema(exportSchema{})
+	if err != nil {
+		return err
+	}
+
+	// only EventStart and EventType should be required
+	for i := 2; i < len(schema); i++ {
```
jerseybbq: This is kinda dirty since it's order dependent...it would be cleaner to iterate through the schema object, and if the field doesn't match EventStart or EventType, then mark if not-required.

warehouse/bigquery.go
```diff
@@ -0,0 +1,394 @@
...
+		return "", err
+	}
+	defer gcsClient.Close()
+
+	_, objName := filepath.Split(filename)
+	w := gcsClient.Bucket(bq.conf.GCS.Bucket).Object(objName).NewWriter(bq.ctx)
+	if _, err = io.Copy(w, f); err != nil {
+		return "", err
+	}
+	if err := w.Close(); err != nil {
```
jerseybbq: I'm not very familiar with the GCS api. Should this be a `defer w.Close()` right after the NewWriter is created?

warehouse/bigquery.go
```diff
@@ -0,0 +1,394 @@
...
+		log.Printf("Could not connect to BigQuery: %s", err)
+		return err
+	}
+	return nil
+}
+
+func (bq *BigQuery) doesTableExist(name string) bool {
+	log.Printf("checking if table %s exists", name)
+
+	q := fmt.Sprintf("SELECT COUNT(*) FROM %s.__TABLES_SUMMARY__ WHERE table_id = '%s';", bq.conf.BigQuery.Dataset, name)
```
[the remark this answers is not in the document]
jameremo: Thanks, that runs much faster as well.

warehouse/bigquery.go
jameremo: It does seem inconsistent with how you would normally handle closing something, but all the samples are written this way, both for creating (https://cloud.google.com/appengine/docs/standard/go/googlecloudstorageclient/app-engine-cloud-storage-sample) and uploading files to GCS (https://github.com/GoogleCloudPlatform/golang-samples/blob/master/storage/gcsupload/gcsupload.go).  

While googling for more about this, I did find one guy who argued that you shouldn't ignore the err resulting from closing a file Writer, so that you can be 100% all your data really did get written to the file (https://joeshaw.org/dont-defer-close-on-writable-files/).  I would imagine the GCS samples are written this way for reasons that are similar in pattern, if not in exact origin.

jameremo:
PTAL @jerseybbq

warehouse/schema.go
```diff
@@ -8,7 +8,7 @@ import (
 	"time"
 )
 
-type exportSchema struct {
+type bundleSchema struct {
```
jerseybbq: Nit: can you add a comment (here or in the README) that explains the difference between the BundleSchema and the ExportTableSchema? It's not intuitive (for someone who wasn't part of our conversation yesterday!), and it's unclear at first glance how `WarehouseField` relates to them. Thanks!

PR #7, opened 2018-01-04
Add note for go requirement

sanjaybv:
Apparently, users are required to use go 1.7 or higher because "context" package is unavailable in versions < 1.7

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 `hauser` is a service to download FullStory data export files and load them into a data warehouse. (Redshift and BigQuery are the only warehouses supported currently. Others are easy to add -- pull requests welcome.)
 
 ## Quick Start
-
+**Note: Requires go 1.7 or higher**
 * Build it (for EC2, for example): ``GOOS=linux GOARCH=amd64 go get github.com/fullstorydev/hauser``
 * Copy the included `example-config.toml` file and customize it for your environment, including your FullStory API key, warehouse host, and credentials. AWS credentials (for S3) come from your local environment.
 * Run it: `./hauser -c <your updated config file>`
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 `hauser` is a service to download FullStory data export files and load them into a data warehouse. (Redshift and BigQuery are the only warehouses supported currently. Others are easy to add -- pull requests welcome.)
 
 ## Quick Start
-**Note: Requires go 1.7 or higher**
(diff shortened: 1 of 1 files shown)


sanjaybv:
@jameremo Can you think of other requirements that would be good to add? if there are more, it's probably better to have them in their own sub-heading.

jhump:
I think you could also just change `warehouse/bigquery.go` and `warehouse/redshift.go` to import "golang.org/x/net/context" instead of "context". Not sure whether we *want* to support Go 1.6 or not, but it seems like all of the deps (including the bigquery and redshift client libraries) support it.

jameremo:
Everything besides go and git should be included in the vendor libraries that are checked into the repo.  I don't think we need to put something in there about needing git, that seems like something every engineer should understand?

I don't have a strong opinion on supporting older versions of go, but per their official release page:

> Each major Go release is supported until there are two newer major releases. For example, Go 1.8 is supported until Go 1.10 is released, and Go 1.9 is supported until Go 1.11 is released.

So I think it's reasonable not to spend a lot of effort caring about 1.6 at this point.  

https://golang.org/doc/devel/release.html

jhump:
> So I think it's reasonable not to spend a lot of effort caring about 1.6 at this point.

Fair enough. I think a lot of libraries still try to support it (including grpc and protobuf) because of Google App Engine. Its support for Go 1.8 just went GA a few months ago, so it is likely that not everyone has moved up from 1.6. But since hauser is an app, not a library, it makes sense to not bother with the far

jhump:
Sorry - didn't mean to close this. Clicked the wrong button :/

jameremo:
Ah, that makes sense, thanks.  I did not realize App Engine Go support had such a lag.

README.md
```diff
@@ -3,7 +3,7 @@
 `hauser` is a service to download FullStory data export files and load them into a data warehouse. (Redshift and BigQuery are the only warehouses supported currently. Others are easy to add -- pull requests welcome.)
 
 ## Quick Start
-
+**Note: Requires go 1.7 or higher**
```
jameremo: My personal preference is that this becomes a bullet point in the list below, rather than a note at the top.  So the first bullet point would be:

 * Make sure you have [installed](https://golang.org/doc/install) Go 1.7 or higher.

This is just personal preference, though.

sanjaybv:
I'm not going to change the imports to support go 1.6.

PR #12, opened 2018-07-10
Dockerized Hauser for Google BigQuery

schuylr:
This doesn't work for Amazon RedShift yet, but should serve as a good jumping-off point to get it started

Executable container is available at `quay.io/jewlr/hauser`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM golang:1.7
+RUN apt update && \
+	 apt install -y gettext-base
+
+COPY server-conf/ /server-conf/
+RUN mkdir -p /go/src/github.com/fullstorydev/hauser/ && \
+	 cp /server-conf/entry.sh /usr/local/bin/entry.sh
+
+ADD . /go/src/github.com/fullstorydev/hauser/
+WORKDIR /go/src/github.com/fullstorydev/hauser/
+RUN go build -o /usr/local/bin/hauser .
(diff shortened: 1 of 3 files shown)


jameremo:
Hi Schuyler, I just wanted to take a second to thank you for this PR, and also to let you know that we're not ignoring it. Internally we are using hauser in a Docker container, but we had not published it because it is opinionated/specific to our use case. 

However, it's also become clear that a lot of potential hauser users would benefit from something a bit more "prepackaged", particularly those who are not familiar with Go. We're finishing up some strategic discussions on this and the general future of hauser, and then we'll get back to you on this PR.

joelash:
I'm assuming this is totally NOT the path that FullStory is planning on taking hauser. I worked on a bit more flexible docker image for hauser today, [repo here](https://github.com/aclaimant/docker-hauser). I need to upload the image somewhere and work on a Readme, but essentially it read everything our of environment variables. Use the value in the config with `HAUSER_` prefix.

rdyson:
> This doesn't work for Amazon RedShift yet, but should serve as a good jumping-off point to get it started
> 
> Executable container is available at `quay.io/jewlr/hauser`

@schuylr a quick note to say thanks for posting this, I was able to deploy Hauser on Amazon Elastic Container Service pushing data to Google Cloud Storage and BigQuery using your config with no tweaks necessary. Cheers!

PR #14, opened 2018-07-19
support schema changes for redshift

butanian:
This PR makes it so the export to Redshift works if changes are made to the export table client side, or if new fields are added to the export.

This is part of a larger arc of work which will -
(1) Add support for changing export schema on Redshift
(2) Add support for changing export schema on BigQuery
(3) [Add new fields to the export](https://github.com/fullstorydev/hauser/pull/17)

New fields (3) will only be merged in once (1) and (2) are done and merged into master.

Additional notes for this PR:
I need to update the Hauser documentation and the changes can be best summarized by -
- The application relies on clients not renaming existing hauser exported column names. 
- If we find that a column is missing in an existing export table (either because we added new fields or a client deleted an existing schema column), we execute alter table statements and add in the missing columns.
- If there are extra columns present on the export table, we only populate the columns whose name matches with the export schema. Indirectly we are relying on any other columns on the table to be nullable.


warehouse/redshift.go
```diff
@@ -149,23 +149,52 @@ func (rs *Redshift) LoadToWarehouse(file string, _ ...fullstory.ExportMeta) erro
...
+			for _, alterStmt := range alterTableStmts {
+				// Redshift only allows addition of one column at a time, hence the the alter statements in a loop yuck
+				if _, err = rs.conn.Exec(alterStmt); err != nil {
+					return err
+				}
+			}
+		}
+	} else {
+		// if the export table does not exist we create a one with all the columns we expect!
+		log.Printf("Export table %s doesnot exist! Creating one!", rs.conf.Redshift.ExportTable)
```
jameremo: nit: extra "a" in "create a one"
nit: need space between does and not

warehouse/redshift.go
```diff
@@ -295,3 +325,60 @@ func (rs *Redshift) DoesTableExist(name string) bool {
...
+	var columns []string
+
+	defer rows.Close()
+	for rows.Next(){
+		var column string
+		err = rows.Scan(&column)
+		columns = append(columns, column)
+	}
+	err = rows.Err() // get any error encountered during iteration
+	if err != nil {
```
jameremo: nit: You can combine these 2 lines into a single if statement

warehouse/redshift.go
```diff
@@ -295,3 +325,60 @@ func (rs *Redshift) DoesTableExist(name string) bool {
...
+	var alterTableStmts []string
+	for _, f := range fields {
+		alterTableStmts = append(alterTableStmts, fmt.Sprintf("ALTER TABLE %s ADD COLUMN %s %s;", tableName, f.Name, f.DBType))
+	}
+
+	return alterTableStmts
+}
+
+func (rs *Redshift) getMissingFields(schema Schema, tableColumns []string) []WarehouseField {
+	existingColumns := make(map[string]bool)
```
jameremo: Some people prefer to make a set in go with `map[string]struct{}`, since `struct{}` takes no storage space. But others think that syntax is too unwieldy. I can see either argument, just wanted to make sure you knew about that subtlety.

warehouse/redshift_test.go
```diff
@@ -36,3 +36,44 @@ func TestRedshiftValueToString(t *testing.T) {
...
+		{wh.exportSchema, []string{"Dummy"}, len(schemaHeaders)},
+		// Only one column common between the table columns, and schema, so there are len(schemaHeaders) - 1 missing field
+		{wh.exportSchema, []string{"PageUrl"}, len(schemaHeaders) - 1},
+		// Same as above, but with additional columns in the export table that we don't care about so still 1 missing field
+		{wh.exportSchema, []string{"Dummy Column1", "PageUrl", "Dummy Column2"}, len(schemaHeaders) - 1},
+
+	}
+
+	for _, testCase := range testCases {
+		missingFieldNames := getFieldNames(wh.getMissingFields(testCase.schema, testCase.columns))
```
jameremo: I don't think we have to call `getFieldNames` here, as the test only ever checks the length of the slice? Or were you planning on adding more checks explicitly on the field name values?

warehouse/redshift.go
```diff
@@ -149,23 +149,52 @@ func (rs *Redshift) LoadToWarehouse(file string, _ ...fullstory.ExportMeta) erro
...
+		// make sure all the columns in the csv export exist in the Export table
+		exportTableColumns := rs.getTableColumns(rs.conf.Redshift.ExportTable)
+		missingFields := rs.getMissingFields(rs.exportSchema, exportTableColumns)
+
+		// If some fields are missing from the fsexport table, either we added new fields
+		// or existing expected columns were deleted by the user we add the relevant columns.
+		// Alter the table and add the missing columns.
+		if len(missingFields) > 0 {
+			log.Printf("Found %d missing fields. Adding columns for these fields.", len(missingFields))
+			alterTableStmts := rs.getAlterTableStatements(missingFields, rs.conf.Redshift.ExportTable)
```
jameremo: nitty nit: I would calculate each alter statement separately, as you need it. Feel free to ignore.

warehouse/redshift_test.go
jameremo: I think you have enough cases that this test is ok. If we were to check field names, the test would be brittle, because every time the schema changed, the test would be likely to break. 

I think if I were to build the test from scratch, I would construct a test schema with only enough fields to cover the test cases. Then checking the names of each field in the result wouldn't be a huge task, and you could likely combine a couple of the test cases.

warehouse/redshift.go
jameremo: I would do just a one-off for now, and then look at refactoring when we do the BigQuery implementation, since we expect there to be some significant refactoring there anyway.

warehouse/redshift.go
```diff
@@ -295,3 +325,56 @@ func (rs *Redshift) DoesTableExist(name string) bool {
...
+	return missingFields
+}
+
+func(rs *Redshift) getSchemaHeaders(schema Schema) []string {
+	var headers []string
+	for _, f := range schema {
+		headers = append(headers, f.Name)
+	}
+	return headers
+}
```
jameremo: nit: Please do add the last newline to all source files; that's pretty standard for all code at FS. A lot of times we don't explicitly point them out in code reviews since GitHub does such a good job of it; we just assume you can see it yourself and fix it.

warehouse/redshift.go
```diff
@@ -295,3 +325,56 @@ func (rs *Redshift) DoesTableExist(name string) bool {
...
+
+	// get any error encountered during iteration
+	if err = rows.Err(); err != nil {
+		log.Fatal(err)
+	}
+	return columns
+}
+
+func (rs *Redshift) getAlterTableStatement(f WarehouseField, tableName string) string {
+	return fmt.Sprintf("ALTER TABLE %s ADD COLUMN %s %s;", tableName, f.Name, f.DBType)
```
jameremo: one last nit: Now that this function is one line, it kind of doesn't need to be a function anymore? For the clean code purists this would be a positive for readability's sake, but I think most go coders would be unhappy with the unnecessary overhead of adding an additional frame to the stack.

warehouse/redshift_test.go
butanian: for this PR I will stop at checking just the lengths. In a separate PR, will work on adding the ability to build custom schemas and check field names in the tests.

warehouse/redshift.go
butanian: I actually have a strong preference for doing something like this. Readability is priceless! Happy to change though, if you think the downsides in GO are decent enough to warrant a non fn approach!

warehouse/redshift.go
butanian: Just set up my (new) IDE to add the last newline to all source files, on save.

warehouse/redshift.go
butanian: a weekend later, and after reading some of the comments in the slack channel, since my 'strong' preference is wavering, changing to inline!

PR #17, opened 2018-07-20
Add dev tools data columns to the export

butanian:
This PR adds the following dev tools fields to the export -
LoadDomContentTime     int64
LoadFirstPaintTime           int64
LoadEventTime                  int64

### DIFF
--- a/warehouse/schema.go
+++ b/warehouse/schema.go
@@ -39,6 +39,9 @@ type bundleEvent struct {
 	UserDisplayName        string
 	UserId                 int64
 	CustomVars             string
+	LoadDomContentTime     int64
+	LoadFirstPaintTime     int64
+	LoadEventTime          int64
 }
 
 // syncTable represents all the fields that should appear in the table used to track which bundles have been synced.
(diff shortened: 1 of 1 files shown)


warehouse/schema.go
```diff
@@ -38,10 +38,10 @@ type bundleEvent struct {
 	UserEmail              string
 	UserDisplayName        string
 	UserId                 int64
-	CustomVars             string
 	LoadDomContentTime     int64
 	LoadFirstPaintTime     int64
 	LoadEventTime          int64
+	CustomVars             string
```
butanian: I had to do this because when we transform the json to csv we hold back all the 'unknown' columns and club them all together and insert as CustomVars in the last column. Added a note so others are not confused as well! We are actively using the schema to map the csv columns to the fsexport table columns.

jameremo:
The primary changes that are in this PR appear to have already been merged in as part of #24, likely related to all the branching issues earlier this week. Let's talk about this one f2f tomorrow.

PR #18, opened 2018-07-24
Add support for schema changes to BigQuery

butanian:
This PR adds BigQuery support for schema changes. If hauser fields are not present in the existing export table in the specified dataset, we now add them and make sure all hauser fields are also present on any existing dataset table.

This is part (2) of a larger arc of work which will -
(1) [Add support for changing export schema on Redshift](https://github.com/fullstorydev/hauser/pull/14)
(2) Add support for changing export schema on BigQuery
(3) [Add new fields to the export](https://github.com/fullstorydev/hauser/pull/17)

### DIFF
--- a/warehouse/bigquery.go
+++ b/warehouse/bigquery.go
@@ -225,7 +225,7 @@ func (bq *BigQuery) GetMissingFields(hauserSchema, bqSchema bigquery.Schema) []*
 	bqSchemaMap := makeSchemaMap(bqSchema)
 	var missingFields []*bigquery.FieldSchema
 	for _, f := range hauserSchema {
-		if _, ok := bqSchemaMap[f.Name]; !ok {
+		if _, ok := bqSchemaMap[strings.ToLower(f.Name)]; !ok {
 			missingFields = append(missingFields, f)
 		}
 	}
@@ -246,7 +246,7 @@ func (bq *BigQuery) AppendToSchema(schema bigquery.Schema, missingFields []*bigq
 func makeSchemaMap(schema bigquery.Schema) map[string]struct{} {
 	schemaMap := make(map[string]struct{})
 	for _, f := range schema {
-		schemaMap[f.Name] = struct{}{}
+		schemaMap[strings.ToLower(f.Name)] = struct{}{}
 	}
 	return schemaMap
 }
--- a/warehouse/bigquery.go
+++ b/warehouse/bigquery.go
@@ -137,10 +137,8 @@ func (bq *BigQuery) LoadToWarehouse(filename string, bundles ...fullstory.Export
 
 	defer bq.deleteFromGCS(objName)
 
-	if !bq.doesTableExist(bq.conf.BigQuery.ExportTable) {
-		if err := bq.createExportTable(); err != nil {
-			return err
-		}
+	if err := bq.EnsureCorrectExportTable(); err != nil {
+		return err
 	}
 
 	// create loader to load from file into export table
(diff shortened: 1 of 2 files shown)


jameremo:
cc @sanjaybv

warehouse/bigquery.go
```diff
@@ -170,6 +168,89 @@ func (bq *BigQuery) LoadToWarehouse(filename string, bundles ...fullstory.Export
...
+	}
+	return nil
+}
+
+// GetMissingFields returns all fields that are present in the hauserSchema, but not in the bqSchema
+func (bq *BigQuery) GetMissingFields(hauserSchema, bqSchema bigquery.Schema) []*bigquery.FieldSchema {
+	bqSchemaMap := makeSchemaMap(bqSchema)
+	var missingFields []*bigquery.FieldSchema
+	for _, f := range hauserSchema {
+		if _, ok := bqSchemaMap[f.Name]; !ok {
```
jameremo: If I'm reading [this doc](https://cloud.google.com/bigquery/docs/schemas#column_names) correctly, BQ column names are also case-insensitive. Interestingly, BQ does appear to consistently keep whatever case was used when creating the table, but I think we still need to handle a possible case difference.

butanian: good catch! Updated the column comparisons to be case insensitive!

butanian:
I definitely want to combine logic between the redshift and bigquery implementations. Maybe even make a generic way that can be applied to other potential new implementations! There are some subtle differences between the two, you pointed one out, and the other is the Schema objects the two implementations use are different. I was hoping to discuss it some and then potentially do it in a different PR since it doesn't affect behavior and you know, how big a fan I am of granular PRs 😬

warehouse/bigquery_test.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	hauserSchema := bigquery.Schema([]*bigquery.FieldSchema {
+		&requiredField1,
+		&requiredField2,
+	})
+
+	var testCases = []struct {
+		hauserSchema	bigquery.Schema
+		tableSchema		bigquery.Schema
```
jameremo: nit: whitespace inconsistency, may be worth it to run `goimports` on this file. 

Last year when I first started doing go, I tried to hook this into my IDE, but it would get into weird reload from disk loops. What was then Gogland is now GoLand, and might be a bit better. I haven't tried it again, largely because of the tools already available in our other repos.

warehouse/bigquery.go
```diff
@@ -170,6 +168,89 @@ func (bq *BigQuery) LoadToWarehouse(filename string, bundles ...fullstory.Export
...
+	}
+	return missingFields
+}
+
+// AppendToSchema adds all the missingFields to the schema.
+// It marks the Required field as false so the change can apply to existing tables with data.
+func (bq *BigQuery) AppendToSchema(schema bigquery.Schema, missingFields []*bigquery.FieldSchema) bigquery.Schema {
+	for _, f := range missingFields {
+		// Need to set required to false since all the older rows will not have any data for the columns we are about to add
+		f.Required = false
```
jameremo: You could move this line up to the `GetMissingFields` function. Then you can replace the entire `AppendToSchema` function with a single line that makes use of Go's spread operator and the fact that `append` is a variadic function:

`md.Schema = append(md.Schema, missingFields...)`

warehouse/bigquery_test.go
jhump: For this, you want `gofmt -w -s`. BTW, I would highly recommend adding a CI build to this repo. Take a look at the `Makefile` and `.travis.yml` files in some of our other repos, like this one:
https://github.com/fullstorydev/grpcurl
(Obviously that should be separate PR -- maybe multiple PRs if you also need to change a decent bit of code to actually make all of the static checks pass.)

warehouse/bigquery.go
sanjaybv: +1

butanian: I have to mark the Required property of the field to false as well. Is there a way to do that as well in addition to the append? It was the reason I created the function.

butanian: done

PR #22, opened 2018-08-06
Add config options that let users upload json files only

butanian:
Separate PR for creating new properties, that lets users upload files as JSON instead of CSV if they are not loading into the warehouse and want to query the files directly.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -30,10 +30,11 @@ type Config struct {
 }
 
 type S3Config struct {
-	Bucket  string
-	Region  string
-	Timeout duration
-	S3Only  bool
+	Bucket              string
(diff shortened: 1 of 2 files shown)


sanjaybv:
Since we don't do "releases", I feel like it is not a good idea to merge this without implementing the code that use the `UploadAsJSON*` fields. It will leave the master in a state where these config options will not work.

butanian:
@sanjaybv 
Definitely not going to merge this unless with the accompanying change. Just a separate PR to make it easy to review.

butanian:
Dropping this PR. If we end up going down the route of supporting JSON only uploads will revisit.

PR #23, opened 2018-08-08
Update Readme

butanian:
anglais is hard

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,16 +11,11 @@
 ## How It Works
 When first run, `hauser` will query FullStory's [data export API](http://help.fullstory.com/develop-rest) to find the earliest export file available. `hauser` will then download all available export files, performing some light transformation for [custom user vars](http://help.fullstory.com/develop-js/setuservars?from_search=17717406) before loading it into the warehouse.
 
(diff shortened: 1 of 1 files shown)


jameremo:
Alors écrivons la documentation en français!

README.md
```diff
@@ -11,12 +11,17 @@
...
 
 Export files may be processed one at a time, or they may be grouped into batches by day using the boolean config option `GroupFilesByDay`.  When grouping is enabled, export files are still processed serially, but all files having the same date (in UTC) will be combined into a single file before upload to the target warehouse.  Grouping files is helpful for loading large amounts of historical data, when the total number of load operations might reach some quota.  BigQuery, for example, limits the number of loads per day on a single table to 1000.
 
 `hauser` can safely be stopped and restarted. For Redshift and BigQuery, it uses the `SyncTable` to keep track of what export files have been processed, and will restart from the last known sync point.
 
+## Addition of new fields to the export
+As FullStory adds more features we expose additional fields in our data export. `hauser` automatically deals with the addition of new fields by appending nullable columns to the warehouse export table.
```
jameremo: I think this section and the other new paragraph above should be combined into a single section that generally discusses how schema changes are handled. As part of that discussion, we may also want to talk about how "extra" table columns that are not part of the export bundle are handled.

This section is also on the same level as `How It Works` above, so it's effectively making the Redshift and BQ sections below its children. I would either make it a child of `How It Works`, or move it to just above `Working with Custom Vars`, so it's better separated.

butanian:
PTAL. Created a separate section called 'Schema Changes'

PR #25, opened 2018-08-09
Reduce db calls to get Export Table columns

butanian:
Previously we were querying the warehouse export table, getting the export table columns, for every record we inserted. This commit makes it so we only query the table columns once after any potential schema updates.

### DIFF
--- a/main.go
+++ b/main.go
@@ -305,6 +305,8 @@ func main() {
 		log.Fatal(err)
 	}
 
+	// Note(aneesh): We SHOULD fetch the table columns ONLY after the call to EnsureCompatibleExportTable.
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -306,13 +305,16 @@ func main() {
 		log.Fatal(err)
 	}
 
+	// Note(aneesh): We SHOULD fetch the table columns ONLY after the call to EnsureCompatibleExportTable.
```
jameremo: nit: We don't usually put names on notes, as they don't imply any further responsibility. We usually do "NB: note text", NB being short for "nota bene".

PR #26, opened 2018-08-09
Respect 'Retry-After' property if defined in the 429 response

butanian:
This change makes it so if we encounter a 429 response when fetching export data, we parse the Retry-After property and respect the specified wait time before retrying.

A few constants were arbitrarily defined - # of retries and default wait Retry-After time if none is specified in the response. If you think there is value in having these constants sourced from the config, happy to move them there.

Changes were made to the vendor/github.com/nishanths/fullstory package. If the changes are acceptable here, I will make corresponding changes in the vendor package and reconcile.

### DIFF
--- a/main.go
+++ b/main.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
@@ -22,6 +23,13 @@ var (
 	conf               *config.Config
 	currentBackoffStep = uint(0)
 	bundleFieldsMap    = warehouse.BundleFields()
+)
+
+const (
+	maxAttempts int = 3
+
+	// Default retry after duration. Arbitrarily set to 10s.
+	defaultRetryAfterDuration time.Duration = time.Duration(10) * time.Second
 )
 
 // Record represents a single export row in the export file
@@ -202,9 +210,30 @@ func LoadBundles (wh warehouse.Warehouse, filename string, bundles ...fullstory.
 	return nil
 }
 
+func getExportData(fs *fullstory.Client, bundleID int) (fullstory.ExportData, error) {
+	log.Printf("Getting Export Data for bundle %d\n", bundleID)
+	for r := 1; r <= maxAttempts; r++ {
+		stream, err := fs.ExportData(bundleID)
+		if err != nil {
+			log.Printf("Failed to fetch export data for Bundle %d", bundleID)
+			retryAfterDuration := defaultRetryAfterDuration
+			if statusError, ok := err.(fullstory.StatusError); ok {
+				if retryAfter, err := strconv.Atoi(statusError.RetryAfter); err == nil {
+					retryAfterDuration = time.Duration(retryAfter) * time.Second
+				}
+			}
+			log.Printf("Attempt #%d failed. Retrying after %s\n", r, retryAfterDuration)
+			time.Sleep(retryAfterDuration)
+		} else {
+			return stream, nil
+		}
+	}
+	return nil, fmt.Errorf("Unable to fetch export data. Tried %d times.", maxAttempts)
+}
+
 // WriteBundleToCSV writes the bundle corresponding to the given bundleID to the csv Writer
 func WriteBundleToCSV(fs *fullstory.Client, bundleID int, csvOut *csv.Writer, wh warehouse.Warehouse) (numRecords int, err error) {
-	stream, err := fs.ExportData(bundleID)
+	stream, err := getExportData(fs, bundleID)
 	if err != nil {
 		log.Printf("Failed to fetch bundle %d: %s", bundleID, err)
 		return 0, err
--- a/main.go
+++ b/main.go
@@ -220,19 +220,20 @@ func getExportData(fs *fullstory.Client, bundleID int) (fullstory.ExportData, er
 		}
 
 		log.Printf("Failed to fetch export data for Bundle %d", bundleID)
+		retryAfterDuration := defaultRetryAfterDuration
 		if statusError, ok := err.(fullstory.StatusError); ok {
 			// If the status code is NOT 429 and the code is below 500 we will not attempt to retry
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -202,9 +210,30 @@ func LoadBundles (wh warehouse.Warehouse, filename string, bundles ...fullstory.
...
+		if err != nil {
+			log.Printf("Failed to fetch export data for Bundle %d", bundleID)
+			retryAfterDuration := defaultRetryAfterDuration
+			if statusError, ok := err.(fullstory.StatusError); ok {
+				if retryAfter, err := strconv.Atoi(statusError.RetryAfter); err == nil {
+					retryAfterDuration = time.Duration(retryAfter) * time.Second
+				}
+			}
+			log.Printf("Attempt #%d failed. Retrying after %s\n", r, retryAfterDuration)
+			time.Sleep(retryAfterDuration)
```
sanjaybv: Hauser may receive errors with non-5XX status codes. In which case, retrying will not help. 5XX error codes are returned when something went wrong on the server side and the request can be tried again in some time.

So, I guess this case, we can check if a `Retry-After` header exists and wait for the specified time. If there was no such header, check if it was a `5XX` error and retry after the default time. Exit early for all other errors.

main.go
```diff
@@ -24,6 +25,13 @@ var (
 	bundleFieldsMap    = warehouse.BundleFields()
 )
 
+const (
+	maxAttempts int = 3
```
sanjaybv: Just a NIT: Hauser makes different kinds of API calls, to BigQuery, RedShift, FullStory etc. I think it would help to be clear to say when exactly these constants will be used. 

So, maybe `maxRetryAttempts` would be more apt along with a comment that says where it will be used (to make fullstory fetch calls).

main.go
```diff
@@ -24,6 +25,13 @@ var (
...
+
+	// Default retry after duration. Arbitrarily set to 10s.
```
sanjaybv: Same NIT about making the comment more descriptive.

vendor/github.com/nishanths/fullstory/client.go
```diff
@@ -65,11 +66,17 @@ func (c *Client) doReq(req *http.Request) (io.ReadCloser, error) {
...
 		io.Copy(b, resp.Body) // Ignore error.
-		return nil, StatusError{
+		statusError := StatusError {
 			Body:       b,
 			Status:     resp.Status,
 			StatusCode: resp.StatusCode,
 		}
+
+		if resp.StatusCode == http.StatusTooManyRequests {
+			statusError.RetryAfter = resp.Header.Get("Retry-After")
```
sanjaybv: `Retry-After` header can be set even if it is not a `http.StatusTooManyRequests`. I believe status `503` may also come with a `Retry-After` header. So you can probably just do this:
```go
return nil, StatusError{
  Body:       b,
  Status:     resp.Status,
  RetryAfter: resp.Header.Get("Retry-After"),
  StatusCode: resp.StatusCode,
}

**Comment by butanian** [comment]
PTAL

File: main.go [hunk 5]
```diff
@@ -201,9 +211,35 @@ func LoadBundles (wh warehouse.Warehouse, filename string, bundles ...fullstory.
...
+				break
+			}
+
+			retryAfterDuration := defaultRetryAfterDuration
+			if retryAfter, err := strconv.Atoi(statusError.RetryAfter); err == nil {
+				retryAfterDuration = time.Duration(retryAfter) * time.Second
+			}
+			log.Printf("Attempt #%d failed. Retrying after %s\n", r, retryAfterDuration)
+			time.Sleep(retryAfterDuration)
+		}
```
jameremo: I guess we intend to retry if `ok` is false, i.e. `err` is not a `fullstory.StatusError`? I think such an error might occur for example if the request never made it to the FS server due to temporary network issues, so it's correct to retry it. 

Right now if I'm reading this right, the retry does occur, but right away. Maybe it's better to put the init logic for `retryAfterDuration` and for the call to `time.Sleep` outside of the check to see whether this is a statusError, so we always retry on early failures, but after the default wait.

butanian: good point. done. Also any thoughts on the 10s default delay between retries, now that we are re-trying a larger set of errors?

jameremo: I think 10s is fine. Worst case scenario, it fails after 3 tries and then bubbles the error up to the main algorithm, which waits 30s before trying again after errors. Effectively this just means some errors that occurred while fetching bundles will be retried a bit faster (10s vs 30s).

main.go
```diff
@@ -201,9 +211,36 @@ func LoadBundles (wh warehouse.Warehouse, filename string, bundles ...fullstory.
...
+			if statusError.StatusCode != http.StatusTooManyRequests && statusError.StatusCode < 500 {
+				break
+			}
+
+			if retryAfter, err := strconv.Atoi(statusError.RetryAfter); err == nil {
+				retryAfterDuration = time.Duration(retryAfter) * time.Second
+			}
+		}
+
+		log.Printf("Attempt #%d failed. Retrying after %s\n", r, retryAfterDuration)
```
jameremo: My apologies, I should have noticed this before, but this function never logs or returns the original error that occurred in the call to `fs.ExportData`. Maybe this line is a good spot to log it? 

In addition to logging the error on each attempt, I think it would be good to return the last error that occurred, or to wrap it.  You can do this with https://godoc.org/github.com/pkg/errors#example-Wrap. It looks like we already have`github.com/pkg/errors` in our dependencies.

butanian: good point. It would be very frustrating for people to not see client errors! done!

sanjaybv:
@butanian @jameremo We shouldn't have merged this PR with changes in the vendor dir, right? It was my understanding that we'd be merging this one after the changes to Nishanth's package was merged.  I guess if we do this, we'll just have to use the "reconcile" option in `gorepoman` and dirty merge it when the PR in Nishanth's package does get merged.

butanian:
@jameremo and I had discussed this. We decided to first change it in the hauser project and then reconcile with gorepoman at a later time.

PR #28, opened 2018-08-15
Update fullstory dep and make corresponding changes

jameremo:
The PR incorporates the changes from https://github.com/nishanths/fullstory/pull/1 into `hauser` via `gorepoman`. It also includes a few small code changes around the fact that the retry wait is now expressed as a `time.Duration` rather than an `int`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -10,7 +10,6 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"time"
 
@@ -223,16 +222,9 @@ func getExportData(fs *fullstory.Client, bundleID int) (fullstory.ExportData, er
 		log.Printf("Failed to fetch export data for Bundle %d: %s", bundleID, err)
 
 		fsErr = err
-		retryAfterDuration := defaultRetryAfterDuration
-		if statusError, ok := err.(fullstory.StatusError); ok {
-			// If the status code is NOT 429 and the code is below 500 we will not attempt to retry
-			if statusError.StatusCode != http.StatusTooManyRequests && statusError.StatusCode < 500 {
-				break
-			}
-
-			if retryAfter, err := strconv.Atoi(statusError.RetryAfter); err == nil {
-				retryAfterDuration = time.Duration(retryAfter) * time.Second
-			}
+		doRetry, retryAfterDuration := getRetryInfo(err)
+		if !doRetry {
+			break
 		}
 
 		log.Printf("Attempt #%d failed. Retrying after %s\n", r, retryAfterDuration)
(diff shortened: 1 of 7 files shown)


main_test.go
```diff
@@ -0,0 +1,48 @@
...
+			expDoRetry:    true,
+			expRetryAfter: defaultRetryAfterDuration,
+		},
+		{
+			err:           fullstory.StatusError{StatusCode: http.StatusTooManyRequests, RetryAfter: 3 * time.Second},
+			expDoRetry:    true,
+			expRetryAfter: 3 * time.Second,
+		},
+		{
+			err:           fullstory.StatusError{StatusCode: http.StatusInternalServerError, RetryAfter: 3 * time.Second},
```
butanian: can you change the status code for this test to StatusServiceUnavailable? It is the 5xx status that actually returns a retry-after. Protects in case someone in the future tries to make the retry logic target more specific 5xx errors.

main.go
```diff
@@ -242,6 +234,21 @@ func getExportData(fs *fullstory.Client, bundleID int) (fullstory.ExportData, er
 	return nil, errors.Wrap(fsErr, fmt.Sprintf("Unable to fetch export data. Tried %d times.", maxAttempts))
 }
 
+func getRetryInfo(err error) (bool, time.Duration) {
+	if statusError, ok := err.(fullstory.StatusError); ok {
+		// If the status code is NOT 429 and the code is below 500 we will not attempt to retry
+		if statusError.StatusCode != http.StatusTooManyRequests && statusError.StatusCode < 500 {
+			return false, defaultRetryAfterDuration
```
butanian: wondering if the retry after we return here should be a stupid long duration or something so people don't by mistake hose fullstory servers if there are client errors?

main_test.go
jameremo: Actually the intent here is to retry on all 5xx errors, and when there is no `Retry-After` header, to use our default duration, regardless of the specific code. The general argument is that you don't know enough about the error (even a 500) to say that retrying is a lost cause, so for the sake of robustness you should give it a go. I've seen cases where some of our internal microservices returned a 500 because of some temporary blip in GCP that we didn't interpret appropriately; we just categorized it as a failure and called it a day.

main.go
jameremo: I guess define "stupid long"? Right now the default wait is 10s, and it doesn't seem like a call every 10s is often enough to hose anything. Keep in mind that without this retry mechanism, the error would bubble up to the main function, which would then wait 30s and try again anyway. So we're not drastically changing the overall retry rate when the error is something "catastrophic" and can't be resolved by waiting.

main_test.go
butanian: Agreed that we should retry all 5xx errors. Its just that 503 errors are most likely to specify a custom retry-after header, hence the recommendation. The 503 tests 5xx errors get retried and also protects against people in the future erroneously ignoring 503 retry-after directives!

main.go
butanian: I guess I said stupid long because I couldn't define it well :p 

agree with your thoughts!

jameremo:
@sanjaybv Ping!

main_test.go
```diff
@@ -0,0 +1,53 @@
+package main
+
+import (
+	"testing"
+	"net/http"
+	"time"
+	"github.com/pkg/errors"
+	"github.com/nishanths/fullstory"
```
sanjaybv: Just a nit; I think the `github.com/*` packages should be in their own block.

main.go
```diff
@@ -223,16 +222,9 @@ func getExportData(fs *fullstory.Client, bundleID int) (fullstory.ExportData, er
...
-
-			if retryAfter, err := strconv.Atoi(statusError.RetryAfter); err == nil {
-				retryAfterDuration = time.Duration(retryAfter) * time.Second
-			}
+		doRetry, retryAfterDuration := getRetryInfo(err)
+		if !doRetry {
+			break
 		}
 
 		log.Printf("Attempt #%d failed. Retrying after %s\n", r, retryAfterDuration)
```
sanjaybv: This can be a TODO; should we maybe log the error here instead of assuming that all errors from one set of `ExportData` calls will always be the same?

jameremo: The `err` is logged separately on line 222 above, and then the very last one to occur is returned below. Does that address your concern? Ultimately, if all attempts fail, the last attempt will be logged twice, once on line 222 and then again wherever this bubbles up in the main algorithm.

sanjaybv: Oops! Sorry, I totally missed the log on line 222. LGTM.

PR #29, opened 2018-08-24
Fix custom var handling in JSON to CSV conversion

jameremo:
There was a tiny mistake in the conversion of each export record from JSON to CSV format. Fixed that mistake and added a test to catch this moving forward.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -4,9 +4,12 @@ import (
 	"testing"
 	"net/http"
 	"time"
-	
+
 	"github.com/pkg/errors"
 	"github.com/nishanths/fullstory"
+	"github.com/fullstorydev/hauser/warehouse"
+	"fmt"
+	"encoding/json"
 )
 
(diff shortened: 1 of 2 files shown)


main_test.go
```diff
@@ -52,3 +55,112 @@ func TestGetRetryInfo(t *testing.T) {
...
+	if str1 == str2 {
+		return true
+	}
+	if len(str1) > 0 && str1[0] == '{' {
+		return compareJSONStrings(t, str2, str2)
+	}
+	return false
+}
+
+func compareJSONStrings(t *testing.T, str1, str2 string) bool {
```
sanjaybv: I think this is fine for now, but just FYI, we have a `fs/fsjson` package with `Equals()` function.

jameremo: Is that a package there we're keeping private to the primary repo? Remember this is the public hauser repo.

PR #30, opened 2018-08-27
Preserve case in CustomVars JSON doc

jameremo:
In previous PRs, we added case-insensitive matching of columns in the target warehouse with values from the incoming export bundle's JSON document. When we did that, we failed to preserve the case of values in the CustomVars JSON document, which is important for any downstream ETL processes. For example, let's say we originally were producing CustomVars that looked like:

`{ "importantValue_str": "Important Data!!", "anImportantNum_real": 41 }`

With the bug, the keys in CustomVars became all lowercase: 

`{ "importantvalue_str": "Important Data!!", "animportantnum_real": 41 }`

This PR returns hauser to the state where case is preserved in CustomVars, while still doing case-insensitive matching to column names in the target warehouse.

### DIFF
--- a/main.go
+++ b/main.go
@@ -44,14 +44,18 @@ type ExportProcessor func(warehouse.Warehouse, []string, *fullstory.Client, []fu
 // For existing export table fields that do not exist in the json record, an empty string is populated.
 func TransformExportJSONRecord(wh warehouse.Warehouse, tableColumns []string, rec map[string]interface{}) ([]string, error) {
 	var line []string
-	// Change all record keys to lower case. We do this because columns are case insensitive for most warehouse solutions.
-	rec = getRecordWithLowerCaseKeys(rec)
-
-	// Map of CustomVars
+	lowerRec := make(map[string]interface{})
 	customVarsMap := make(map[string]interface{})
+
+	// Do a single pass over the data record to:
+	// a) extract all custom variables
+	// b) change standard/non-custom field names to lowercase for case-insensitive column name matching
 	for key, val := range rec {
-		if _, ok := bundleFieldsMap[key]; !ok {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -44,14 +44,18 @@ type ExportProcessor func(warehouse.Warehouse, []string, *fullstory.Client, []fu
...
-	// Change all record keys to lower case. We do this because columns are case insensitive for most warehouse solutions.
-	rec = getRecordWithLowerCaseKeys(rec)
-
-	// Map of CustomVars
+	lowerRec := make(map[string]interface{})
 	customVarsMap := make(map[string]interface{})
+
+	// Do a single pass over the data record to:
+	// a) extract all custom variables
+	// b) change standard/non-custom field names to lowercase for case-insensitive column name matching
```
jameremo: You could argue here that we lose some readability by removing the `getRecordWithLowerCaseKeys` function and doing everything in one pass.  However, `TransformExportJSONRecord` is a function that will be called on EVERY record in the bundle. This is a case where I think a slight loss of readability is an ok trade-off for saving those iterations.

jameremo:
About making sure table columns are lowercase: The BQ implementation of `Warehouse.GetExportTableColumns` does forcibly make things lowercase. For Redshift, I read through these two docs:

https://docs.aws.amazon.com/redshift/latest/dg/r_names.html
https://aws.amazon.com/about-aws/whats-new/2017/10/amazon-redshift-announces-support-for-uppercase-column-names/

There is a session-based parameter that you can optionally set to return column names in all uppercase. It doesn't appear from these docs that the Redshift practice of folding column names to lowercase is going to change any time soon, so I don't think it's urgent to add logic to force the column names to lowercase in hauser.

PR #33, opened 2018-11-30
Add EventCustomName to hauser

donjaime:
Add EventCustomName to hauser.

### DIFF
--- a/warehouse/schema.go
+++ b/warehouse/schema.go
@@ -20,7 +20,7 @@ type bundleEvent struct {
 	EventModError          int64
 	EventModSuspicious     int64
 	IndvId                 int64
-	PageClusterId	         int64
+	PageClusterId          int64
 	PageUrl                string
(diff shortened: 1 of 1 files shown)


warehouse/schema.go
```diff
@@ -19,7 +20,7 @@ type bundleEvent struct {
 	EventModError          int64
 	EventModSuspicious     int64
 	IndvId                 int64
-	PageClusterId	       int64
+	PageClusterId	         int64
```
jameremo: nit: weird whitespace

donjaime: This was actually weird in my IDE prior to this. I think we have mixed tabs and spaces. And my tab width is different than ya'lls. Fixing.

donjaime: Ran `go fmt` on the file. We should probably setup some presubmit hooks etc... on these repos.

Issue #34, opened 2018-12-10
S3: allow export files to be stored in subdirectories


patrick-fs:
### Expected behavior

When the `Bucket` parameter in config.toml is set to either

* `somebucket/somesubdirectory`
OR
* `somebucket/somesubdirectory/`

Then the Data Bundle export files should be stored at that path. For example: `somebucket/somesubdirectory/123456.csv`

### Actual Behavior

When the `Bucket` parameter is set to `somebucket/somesubdirectory` in config.toml, the following error is returned:
```
Failed to upload file /tmp/123456.csv to s3: SignatureDoesNotMatch: The request signature 
we calculated does not match the signature you provided. Check your key and signing method.
```
When the `Bucket` parameter is set to `somebucket/somesubdirectory/` in config.toml, the following error is returned:
```
Failed to load file '/tmp/123456.csv' to warehouse: pb: The specified S3 prefix 
somesubdirectory//123456.csv' does not exist
```

patrick-fs:
NOTE: I've already got a fix for this on my fork: https://github.com/patrick-fs/hauser/tree/patrick/s3-name-fix

PR forthcoming

patrick-fs:
PR merged

PR #39, opened 2018-12-12
fix for issue https://github.com/fullstorydev/hauser/issues/34

patrick-fs:
Verified the following values for config.toml/s3/Bucket:

patrick-fs-test/fullstory/
patrick-fs-test/fullstory
patrick-fs-test/
patrick-fs-test

### DIFF
--- a/warehouse/redshift.go
+++ b/warehouse/redshift.go
@@ -110,14 +110,19 @@ func (rs *Redshift) UploadFile(name string) (string, error) {
 	defer cancelFn()
 
 	_, objName := filepath.Split(name)
-	s3path := fmt.Sprintf("s3://%s/%s", rs.conf.S3.Bucket, objName)
+
+	bucketParts := strings.Split(rs.conf.S3.Bucket, "/")
+	bucketName := bucketParts[0]
+	keyPath := strings.Trim(strings.Join(bucketParts[1:], "/"), "/");	
+	key := fmt.Sprintf("%s/%s", keyPath, objName)
 
 	_, err = svc.PutObjectWithContext(ctx, &s3.PutObjectInput{
-		Bucket: aws.String(rs.conf.S3.Bucket),
-		Key:    aws.String(objName),
+		Bucket: aws.String(bucketName),
+		Key:    aws.String(key),
 		Body:   file,
 	})
 
+	s3path := fmt.Sprintf("s3://%s/%s", bucketName, key)
 	return s3path, err
 }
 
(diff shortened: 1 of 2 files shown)


jhump:
@patrick-fs, the CI check that failed was that the code is not properly formatted. Just run `gofmt -w -s  warehouse` in the repo directory and commit and push the resulting change.

patrick-fs:
@jhump yep! added a semi-colon at the end of a statement.

Old habits die hard :)

I've pushed the change.

warehouse/redshift.go
```diff
@@ -110,14 +110,19 @@ func (rs *Redshift) UploadFile(name string) (string, error) {
 	defer cancelFn()
 
 	_, objName := filepath.Split(name)
-	s3path := fmt.Sprintf("s3://%s/%s", rs.conf.S3.Bucket, objName)
+
+	bucketParts := strings.Split(rs.conf.S3.Bucket, "/")
+	bucketName := bucketParts[0]
+	keyPath := strings.Trim(strings.Join(bucketParts[1:], "/"), "/")
+	key := fmt.Sprintf("%s/%s", keyPath, objName)
```
jameremo: What this block is doing is subtle enough that it would be great to have a test for it, both to serve as documentation and to make sure we don't inadvertently break this in the future. You could refactor this logic into a function like so:

```
func getBucketAndKey(bucketPath, objName string) (string, string){
	bucketParts := strings.Split(bucketPath, "/")
	keyPath := strings.Trim(strings.Join(bucketParts[1:], "/"), "/")
	key := fmt.Sprintf("%s/%s", keyPath, objName)
	return bucketParts[0], key
}
```

Then in `redshift_test.go`, add a test function similar to this:

```
func TestGetBucketAndKey(t *testing.T) {
	testCases := []struct {
		s3Config string
		fileName string
		expBucket string
		expKey string
	} {
		{
			s3Config: "plainbucket",
			fileName: "data.csv",
			expBucket: "plainbucket",
			expKey: "data.csv",
		},
		{
			s3Config: "withslash/",
			fileName: "data.csv",
			expBucket: "withslash",
			expKey: "data.csv",
		},
		// TODO: test cases around subdirectories
	}
	
	for i, tc := range testCases {
		// TODO: call getBucketAndKey with tc.sConfig and tc.fileName
		// TODO: compare result with expected values, t.Errorf if they don't match
	}
}
```

patrick-fs: Done!

Issue #38, opened 2018-12-12
fullstory.ExportList is throwing a 404 error for bigquery


sonalcodeeight:
```
2018/12/12 14:09:05 Checking if table fullstory_data exists
2018/12/12 14:09:06 Export table exists, making sure the schema in BigQuery is compatible with the schema specified in Hauser
2018/12/12 14:09:07 Checking if table fullstory_sync_data exists
2018/12/12 14:09:08 Checking if table fullstory_data exists
2018/12/12 14:09:09 Checking for new export files since 0001-01-01 00:00:00 +0000 UTC
2018/12/12 14:09:10 Failed to fetch export list: fullstory: response error: Status:404 Not Found, StatusCode:404, RetryAfter:0
```
What am I missing? 

Here is my config - i've set bucket name, project name, dataset name and table names:
```
FsApiToken = "token"
Backoff = "30s"
BackoffStepsMax = 8
CheckInterval = "30m"
TmpDir = "/tmp"
Warehouse="bigquery"
GroupFilesByDay = false

[gcs]
Bucket = "bucketname"
GCSOnly = false

[bigquery]
Project = "projectname"
Dataset = "test_fullstory"
ExportTable = "fullstory_data"
SyncTable = "fullstory_sync_data"
```

Thanks!

jameremo:
Hello! It looks like hauser could could not find any data export bundles to download. It's using FullStory's [data export API](https://help.fullstory.com/develop-rest/data-export-api) to download data, which requires that you have a valid [API key](https://help.fullstory.com/using/170703-where-can-i-find-my-api-key) in the `FsApiToken` setting. I apologize if you have that set up, it's hard to tell if it's simply been redacted in the config excerpt above. You could maybe try the data export APIs in Postman or some other test REST client just to make sure that you can access them independently of hauser.

Beyond that, it's hard to make a more definite diagnosis without knowing some details about your account. For that, it would be best if you contact support@fullstory.com, and mention this issue so that they route it back to me.

Hope that helps!

sonalcodeeight:
@jameremo  thanks for your response.

I did confirm that i can call `https://export.fullstory.com/api/v1/export/list` with the api token wrapped in the Authorization header as specified by the data export api and I do get a list of data via postman.

But I still get this 404 error with that same api token in the config.

Thanks.

jameremo:
[...]

PR #41, opened 2018-12-17
Add go.mod so hauser can be built using Go 1.11 with modules enabled

### LINKED ISSUE #40: Hauser cloned from git doesn't build outside of a specific folder path
If we clone the hauser repo using "git clone", and then try to build it using "go build", it will fail UNLESS it's inside the folder named .../github.com/fullstorydev/hauser. The error message it produces is:

./main.go:202:30: cannot use bundles (type []"github.com/nishanths/fullstory".ExportMeta) as type []"github.com/fullstorydev/hauser/vendor/github.com/nishanths/fullstory".ExportMeta in argument to wh.LoadToWarehouse
./main.go:211:29: cannot use bundles (type []"github.com/nishanths/fullstory".ExportMeta) as type []"github.com/fullstorydev/hauser/vendor/github.com/nishanths/fullstory".ExportMeta in argument to wh.SaveSyncPoints

The issue can also be fixed by running the "go mod init" command before building ("go mod init" is available starting with Go 1.11)

jhump:
Fixes #40 

Probably easiest to review one commit at a time. Note that I've added `go.mod` files under `vendor` to work-around an issue with modules (that is marked "working as intended"): https://github.com/golang/go/issues/24110


jameremo:
I think this will also fix #38. I did some local testing today, and found that there are some changes to the vendored copy of [nishanths/fullstory](https://github.com/nishanths/fullstory) that cannot be circumvented with config.

jhump:
> Did you just use `go mod init` to generate the `go.mod` file for all the vendored packages?

More or less. I don't have Go 1.11 installed, so I'm actually using `vgo`. Running `vgo mod init` just creates an empty module file. So I then ran `vgo mod tidy` to have it auto-compute the `require` statements for the dependencies.

PR #42, opened 2018-12-18
Greg/localstore

glarchev:
Change to enable hauser to save exports locally without warehousing them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # hauser
 
-`hauser` is a service to download FullStory data export files and load them into a data warehouse. (Redshift and BigQuery are the only warehouses supported currently. Others are easy to add -- pull requests welcome.) There's also an option to save exported files locally.
+`hauser` is a service to download FullStory data export files and load them into a data warehouse. Alternatively, data export files can be saved to local disk. Redshift and BigQuery are the only warehouses supported currently. (Others are easy to add -- pull requests welcome.)
 
 ## Quick Start
 * Make sure you have [installed](https://golang.org/doc/install) Go 1.7 or higher.
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # hauser
 
-`hauser` is a service to download FullStory data export files and load them into a data warehouse. (Redshift and BigQuery are the only warehouses supported currently. Others are easy to add -- pull requests welcome.)
+`hauser` is a service to download FullStory data export files and load them into a data warehouse. (Redshift and BigQuery are the only warehouses supported currently. Others are easy to add -- pull requests welcome.) There's also an option to save exported files locally.
 
 ## Quick Start
 * Make sure you have [installed](https://golang.org/doc/install) Go 1.7 or higher.
@@ -34,6 +34,10 @@ The BigQuery `ExportTable` is expected to be a date partitioned table. As with t
 Loading data into BigQuery may be skipped by setting `GCS.GCSOnly` in the config file to `true`. In this mode, files are copied to GCS, where they remain without being loaded into BigQuery.
 
 If `hauser` detects that a load failure occurred, to ensure data consistency it will roll back all sync points for the most recent date partition and reload all files for the entire partition.
+
+## Local Storage Notes
+
+To only store downloaded export files locally, set the `Warehouse` option to `local`. This will save each export as a json file to the folder specified by `SaveDir` parameter, starting at time specified by `StartTime`
 
 ## Schema Changes
 As FullStory adds more features we expose additional fields in our data export. `hauser` automatically deals with the addition of new fields by appending nullable columns to the warehouse export table.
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -24,6 +25,7 @@ var (
 	conf               *config.Config
 	currentBackoffStep = uint(0)
 	bundleFieldsMap    = warehouse.BundleFields()
+	beginningOfTime = time.Date(2015, 1, 1, 0, 0, 0, 0, time.UTC)
```
butanian: alignment :)

README.md
```diff
@@ -1,6 +1,6 @@
 # hauser
 
-`hauser` is a service to download FullStory data export files and load them into a data warehouse. (Redshift and BigQuery are the only warehouses supported currently. Others are easy to add -- pull requests welcome.)
+`hauser` is a service to download FullStory data export files and load them into a data warehouse. (Redshift and BigQuery are the only warehouses supported currently. Others are easy to add -- pull requests welcome.) There's also an option to save exported files locally.
```
butanian: maybe update the comment so the new feature of saving the bundles locally is not lost? Just felt that could be missed.

@jameremo Thoughts on something like this? 

`hauser` is a service to download FullStory data export files and load them into a data warehouse. Alternatively data export files can be saved to local disk. Redshift and BigQuery are the only warehouses currently supported. (Others are easy to add -- pull requests welcome.)

main.go
```diff
@@ -302,6 +304,58 @@ func WriteBundleToCSV(fs *fullstory.Client, bundleID int, tableColumns []string,
...
+
+		written, err := io.Copy(outfile, gzstream)
+		if err != nil {
+			log.Printf("Failed to copy input stream to file: %s", err)
+			return
+		}
+
+		log.Printf("Copied %d bytes to file", written)
+	}
+
```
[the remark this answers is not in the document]
glarchev: thanks, fixing

example-config.toml
```diff
@@ -39,3 +39,7 @@ Project = "<your project>"
 Dataset = "<your dataset>"
 ExportTable = "<your export table>"
 SyncTable = "<your sync table>"
+
+[local]
```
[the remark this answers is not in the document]
glarchev: I thought about it, but wasn't sure if users would want that. If they do, then we can add that in. Another potential option would be to add the ability to save all bundles for a each day in a file, sort of like what ProcessFilesByDay does now for uploading

main.go
glarchev: thanks, fixing

glarchev:
@butanian I like the idea of letting users upload their local bundles to a warehouse -- if that's a feature they want, we can add that

main.go
```diff
@@ -24,6 +25,7 @@ var (
 	conf               *config.Config
 	currentBackoffStep = uint(0)
 	bundleFieldsMap    = warehouse.BundleFields()
+	beginningOfTime    = time.Date(2015, 1, 1, 0, 0, 0, 0, time.UTC)
```
jameremo: Right now this same constant appears in 2 different packages. It might be better to define it as an exported constant in `schema.go`, from where it can be used both in the main package and in the warehouse package. But that difficulty may go away with my proposal that persisting to local disk be modeled as another warehouse. The code for the `LocalWarehouse` would then reside in the warehouse package, where it would have access to the original constant.

example-config.toml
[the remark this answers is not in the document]
jameremo: In the past, we have had requests to allow for uploading the raw JSON files to S3, as I think there are some tools that can search JSON files, and are in some use cases cheaper than Redshift.

main.go
glarchev: I thought about modeling local store as another warehouse. My concern is that there are 10 different methods that Warehouse interface defines, most of which will probably do nothing for local store.

jameremo: That's fairly usual in situations like this; could you be more specific about the nature of your concern? Is there something that outweighs the consistency and code reuse we would from keeping the same cognitive model?

example-config.toml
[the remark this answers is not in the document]
butanian: @jameremo yes, I remember that now! Its probably what made me think about the save to s3 interface!

I still think its nice to have a standardized interface for both saving to s3 and saving to local disk. But as Jess suggested, we can work on supporting csv/json for warehouses in a separate PR.

main.go
glarchev: I may be mistaken, but looking at the list of Warehouse methods, the only method that would clearly make sense for local store is LoadToWarehouse. (I can see how we could maybe do something interesting with LastSyncPoint and DeleteFile, but even that seems to be a bit of a stretch.) If a class would only implement 1 out of 10 interface methods, is it really a good fit for that interface?

jameremo: You are not mistaken, but there is more argument to be had, and I would prefer that we hash it out f2f. I will put a meeting on the calendar for us tomorrow.

Issue #35, opened 2018-12-10
Redshift: allow user-defined schemas for export and sync table operations


patrick-fs:
### Scenario 1:  "search_path" value provided for schema configuration in config.toml/redshift/DatabaseSchema

Given there is a `DatabaseSchema` configuration parameter in config.toml/redshift
When `DatabaseSchema = ”search_path”`
  and Hauser is executed
Then the tables defined in config.toml/redshift/ExportTable and config.toml/redshift/SyncTable are both created under the schema defined by the database's [Search Path](https://docs.aws.amazon.com/redshift/latest/dg/r_search_path.html) configuration.

### Scenario 2: Non-empty and non "search_path" value provided for schema configuration in config.toml/redshift/DatabaseSchema

Given there is a `DatabaseSchema` configuration parameter in config.toml/redshift
When `DatabaseSchema = ”some_schema”`
  and Hauser is executed
Then the tables defined in config.toml/redshift/ExportTable and config.toml/redshift/SyncTable are both created under the “some_schema” schema

### Scenario 3: No DatabaseSchema parameter provided in config.toml

Given a `DatabaseSchema` configuration parameter is not defined or `DatabaseSchema = ""` in config.toml/redshift
When Hauser is executed
Then throw an error explaining that `DatabaseSchema` must be provided in config.toml/redshift
  and include a link to redshift config documentation in the error message

## Developer Notes

Be aware of backwards compatibility with existing Hauser users. The existing Hauser implementation explicitly creates export and sync tables under the `public` schema. Here's how I think we can support the existing behavior and also effectively communicate the change during runtime:

* To support existing behavior, add a `DatabaseSchema = "public"` entry to example-config.toml/redshift
* Scenario 3 above is intended to communicate new behavior to existing users

patrick-fs:
UPDATE: I changed the first and third scenarios. Whereas I originally wrote Scenario 1 to apply the search_path when `Schema = ""`, I came to the conclusion that this is non-interpretable behavior and decided to make the search_path behavior explicit. I added `Schema = ""` to Scenario 3.

patrick-fs:
PR merged

PR #43, opened 2018-12-26
Patrick/custom redshift schemas

patrick-fs:
Happy Holidays! I hope you and yours are well.

This PR will allow customer-defined schemas to be used when integrating with Redshift: https://github.com/fullstorydev/hauser/issues/35

I've smoke-tested for the "public" schema, "search_path," and an arbitrary schema I created called "patrickschema2"

### DIFF
--- a/warehouse/redshift.go
+++ b/warehouse/redshift.go
@@ -370,8 +370,8 @@ func (rs *Redshift) DoesTableExist(name string) bool {
 	log.Printf("Checking if table %s exists", name)
 
 	var exists int
-	query := fmt.Sprintf("SELECT count(*) FROM information_schema.tables WHERE table_schema = %s AND table_name = '%s';", rs.getSchemaParameter(), name)
-	err := rs.conn.QueryRow(query).Scan(&exists)
+	query := fmt.Sprintf("SELECT count(*) FROM information_schema.tables WHERE table_schema = %s AND table_name = $1;", rs.getSchemaParameter())
+	err := rs.conn.QueryRow(query, name).Scan(&exists)
 	if err != nil {
 		// something is horribly wrong...just give up
 		log.Fatal(err)
@@ -382,8 +382,8 @@ func (rs *Redshift) DoesTableExist(name string) bool {
 func (rs *Redshift) getTableColumns(name string) []string {
 	log.Printf("Fetching columns for table %s", name)
 	ctx := context.Background()
-	query := fmt.Sprintf("SELECT column_name FROM information_schema.columns WHERE table_schema = %s AND table_name  = '%s';", rs.getSchemaParameter(), name)
-	rows, err := rs.conn.QueryContext(ctx, query)
+	query := fmt.Sprintf("SELECT column_name FROM information_schema.columns WHERE table_schema = %s AND table_name  = $1;", rs.getSchemaParameter())
+	rows, err := rs.conn.QueryContext(ctx, query, name)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/warehouse/redshift.go
+++ b/warehouse/redshift.go
@@ -47,26 +47,28 @@ func NewRedshift(c *config.Config) *Redshift {
 	}
 }
 
-func (rs *Redshift) getExportTableName() string {
-	if rs.conf.Redshift.TableSchema == "search_path" {
+func (rs *Redshift) getExportTableName(fullyQualifiedName bool) string {
+	if rs.conf.Redshift.TableSchema == "search_path" || !fullyQualifiedName {
 		return rs.conf.Redshift.ExportTable
 	}
 	return fmt.Sprintf("%s.%s", rs.conf.Redshift.TableSchema, rs.conf.Redshift.ExportTable)
 }
 
-func (rs *Redshift) getSyncTableName() string {
-	if rs.conf.Redshift.TableSchema == "search_path" {
+func (rs *Redshift) getSyncTableName(fullyQualifiedName bool) string {
+	if rs.conf.Redshift.TableSchema == "search_path" || !fullyQualifiedName {
 		return rs.conf.Redshift.SyncTable
 	}
 	return fmt.Sprintf("%s.%s", rs.conf.Redshift.TableSchema, rs.conf.Redshift.SyncTable)
 }
 
-func (rs *Redshift) getSchemaName() string {
+func (rs *Redshift) getSchemaParameter() string {
+	// the built-in current_schema() function will walk the Postgres search_path to get a schema name
+	// more info: https://www.postgresql.org/docs/9.4/functions-info.html
 	if rs.conf.Redshift.TableSchema == "search_path" {
 		return "current_schema()"
 	}
 
-	return fmt.Sprintf("%s", rs.conf.Redshift.TableSchema)
+	return fmt.Sprintf("'%s'", rs.conf.Redshift.TableSchema)
 }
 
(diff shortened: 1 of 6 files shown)


warehouse/redshift.go
```diff
@@ -330,12 +365,13 @@ func (rs *Redshift) RemoveOrphanedRecords(lastSync pq.NullTime) error {
...
 }
 
-// DoesTableExist checks if a table with a given name exists in the public schema
+// DoesTableExist checks if a table with a given name exists
 func (rs *Redshift) DoesTableExist(name string) bool {
 	log.Printf("Checking if table %s exists", name)
 
 	var exists int
-	err := rs.conn.QueryRow("SELECT count(*) FROM pg_tables WHERE schemaname = 'public' AND tablename = $1;", name).Scan(&exists)
+	query := fmt.Sprintf("SELECT count(*) FROM pg_tables WHERE schemaname = %s AND tablename = '%s';", rs.getSchemaParameter(), name)
```
jameremo: nit: Here and in `getTableColumns` below, I think the `$1` query parameter for the table name was preferable (rather than sprintf-ing literals into the SQL string) to prevent SQL injection. Given that the schema may sometimes be a function call, I guess we can't use a parameter for that.

The way we use `hauser` makes it very unlikely of course that there will be any avenues for actual SQL injection, but when it doesn't make the code awkward, I prefer to keep protecting against it. You never know how this code might evolve or who might copy/paste sections, so it's best that it be bullet-proof.

patrick-fs:
UPDATE: I've verified that export and sync tables can be created with arbitrary database schemas (e.g. "patricktest.exportfs"). All scenarios are proven to work in our dev Redshift database.

@butanian @sanjaybv @jameremo

warehouse/redshift.go
```diff
@@ -46,6 +47,37 @@ func NewRedshift(c *config.Config) *Redshift {
 	}
 }
 
+func (rs *Redshift) getExportTableName(fullyQualifiedName bool) string {
```
jameremo: I wonder if it's better to encode the desire for an undecorated vs. a fully qualified table name into calling/not calling a function, rather than in a boolean parameter to this function. In other words, in all the places where you might call `getExportTableName(false)`, instead just directly refer to `rs.conf.Redshift.ExportTable`? You could then maybe rename this function to `qualifiedExportTableName()` (with no params). The intent of calling it (or not calling it) will be clear to the reader of the calling code without having to read this function. It will also simplify the below if statement a bit.

Same for `getSyncTableName` below.

warehouse/redshift_test.go
```diff
@@ -114,3 +122,137 @@ func TestGetBucketAndKey(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range testCases {
+		wh := NewRedshift(tc.conf)
+		err := wh.validateSchemaConfig()
+		if tc.hasError && err == nil {
+			t.Errorf("expected Redshift.validateSchemaConfig() to return an error when config.Config.Redshift.DatabaseSchema is empty")
+		}
+		if tc.hasError && err.Error() != tc.errMessage {
```
jameremo: I think checking for an exact error message makes this a fragile test, and doesn't add much value at the moment since the function only returns one type of error. We're likely to change the error message thinking it has no consequences. Although fixing a unit test is a small consequence, it's a bit of extra work we can avoid.

warehouse/redshift.go
patrick-fs: I totally agree. String concatenation is used to build SQL queries throughout redshift.go (insert, create, and update statements). The two statements that used parametrized SQL were the `DoesTableExist` and `getTableColumns` functions. A deeper refactor is in order to parameterize all queries. Until then, I'll add $1 back to the two statements that I changed.

warehouse/redshift_test.go
patrick-fs: This thought occurred to me as well - I appreciate the general brittleness that comes with testing error message semantics. Having said that: I'd like to keep this one in place since this message is a crucial element to communicating new behavior to existing hauser users (when they decide to upgrade to the latest version). My expectation is that existing users will get the latest version and might very well forget to add the `DatabaseSchema` parameter to their config. They should immediately be informed that hauser won't work this way (and why) and what they can do to fix it.

warehouse/redshift.go
patrick-fs: Great point - I'll fix this up.

warehouse/redshift.go
patrick-fs: In the process of updating this, I found an issue with the query in `getTableColumns` that prevented columns for existing tables using arbitrary schemas from being fetched. The original query used the `PG_TABLE_DEF` table, which will only return tables for schemas that are in the search_path (see https://docs.aws.amazon.com/redshift/latest/dg/r_PG_TABLE_DEF.html). This wasn't apparent when all tables were created using the "public" schema (which is in the search_path by default).

I fixed this by querying system tables under the `information_schema` schema, which is actually a SQL standard: https://www.postgresql.org/docs/9.1/information-schema.html

warehouse/redshift.go
```diff
@@ -46,6 +47,37 @@ func NewRedshift(c *config.Config) *Redshift {
...
+	// more info: https://www.postgresql.org/docs/9.4/functions-info.html
+	if rs.conf.Redshift.DatabaseSchema == "search_path" {
+		return "current_schema()"
+	}
+
+	return fmt.Sprintf("'%s'", rs.conf.Redshift.DatabaseSchema)
+}
+
+func (rs *Redshift) validateSchemaConfig() error {
+	if rs.conf.Redshift.DatabaseSchema == "" {
```
butanian: was wondering if no schema is defined, then relying on current schema is a good default? From what I've understood about the hauser user base, a majority are not power users, and probably just use the default public schema.

patrick-fs: That was my original thought as well, but I came to a different conclusion after further consideration. The behavior of hauser when the DatabaseSchema parameter is empty would not have been clear and could have resulted in putting data in an unintended location. By not accepting a blank parameter and returning a link to documentation when it occurs, we can prevent this type of confusion.

To be sure, using the public schema may not be an issue for some businesses, but our enterprise customers have lately expressed a need to have more control over where the export data gets migrated to in their data warehouse.

patrick-fs: I should also mention: I updated example-config.toml with `DatabaseSchema = "public"`

New users will likely start with this file when creating their own configs, so they would expect the public schema to be used unless the intentionally changed the config to use a different schema.

butanian: SGTM!

warehouse/redshift_test.go
jameremo: I agree that the error message is important, but still disagree that its exact content needs testing. This is already merged, but I feel pretty strongly on this subject, and we can maybe hash it out f2f if this comes up in future PRs.

warehouse/redshift.go
jameremo: Nice, thanks!

warehouse/redshift_test.go
patrick-fs: Hi Jess,

Understood - I've got another PR coming to verify the invocation of ValidateDatabaseSchema + refactor validation into config. The branch is out, I just haven't made the PR yet (I wanted to stew on it until Wednesday). We can discuss the error condition test and make a change as part of that PR if this works for you.

Thank you,

-Patrick

PR #45, opened 2019-01-02
Greg/localstore2

glarchev:
Reworked code to save export files to local storage -- such that localdisk now implements the warehouse interface. Also pulled out some of the common ValueToString functionality into the warehouse.go file, and added the code to save to json to main.go.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -27,6 +27,9 @@ type Config struct {
 	// gcloud: GCS + BigQuery
 	GCS      GCSConfig
 	BigQuery BigQueryConfig
+
+	// local filesystem: Local
+	Local    LocalConfig
 }
 
 type S3Config struct {
@@ -63,6 +66,11 @@ type BigQueryConfig struct {
 
 type duration struct {
 	time.Duration
+}
+
+type LocalConfig struct {
+	SaveDir		string
+	StartTime	time.Time
 }
 
 func (d *duration) UnmarshalText(text []byte) error {
--- a/config/config.go
+++ b/config/config.go
@@ -69,8 +69,10 @@ type duration struct {
 }
 
 type LocalConfig struct {
-	SaveDir   string
-	StartTime time.Time
+	SaveDir      string
+	StartTime    time.Time
+	UseStartTime bool
+	SaveAsJson   bool
 }
 
 func (d *duration) UnmarshalText(text []byte) error {
--- a/config/config.go
+++ b/config/config.go
@@ -29,7 +29,7 @@ type Config struct {
 	BigQuery BigQueryConfig
 
 	// local filesystem: Local
-	Local    LocalConfig
+	Local LocalConfig
 }
 
 type S3Config struct {
@@ -69,8 +69,8 @@ type duration struct {
 }
 
 type LocalConfig struct {
-	SaveDir		string
-	StartTime	time.Time
+	SaveDir   string
+	StartTime time.Time
 }
 
 func (d *duration) UnmarshalText(text []byte) error {
--- a/main.go
+++ b/main.go
@@ -324,7 +324,7 @@ func SaveToLocalFile(conf *config.Config) {
 
 	for _, e := range exports {
 		log.Printf("Processing bundle %d (start: %s, end: %s)", e.ID, e.Start.UTC(), e.Stop.UTC())
-		filename := filepath.Join(conf.Local.SaveDir, fmt.Sprintf("%d.gz", e.ID))
+		filename := filepath.Join(conf.Local.SaveDir, fmt.Sprintf("%d.json", e.ID))
 		outfile, err := os.Create(filename)
 		if err != nil {
 			log.Printf("Failed to create tmp file: %s", err)
--- a/main.go
+++ b/main.go
@@ -25,7 +25,6 @@ var (
 	conf               *config.Config
 	currentBackoffStep = uint(0)
 	bundleFieldsMap    = warehouse.BundleFields()
-	beginningOfTime    = time.Date(2015, 1, 1, 0, 0, 0, 0, time.UTC)
 )
 
 const (
@@ -109,28 +108,41 @@ func ProcessExportsSince(wh warehouse.Warehouse, tableColumns []string, since ti
 func ProcessFilesIndividually(wh warehouse.Warehouse, tableColumns []string, fs *fullstory.Client, exports []fullstory.ExportMeta) (int, error) {
 	for _, e := range exports {
 		log.Printf("Processing bundle %d (start: %s, end: %s)", e.ID, e.Start.UTC(), e.Stop.UTC())
-		filename := filepath.Join(conf.TmpDir, fmt.Sprintf("%d.csv", e.ID))
 		mark := time.Now()
-		outfile, err := os.Create(filename)
-		if err != nil {
-			log.Printf("Failed to create tmp file: %s", err)
-			return 0, err
-		}
-		defer os.Remove(filename)
-		defer outfile.Close()
-		csvOut := csv.NewWriter(outfile)
+		var filename string
+		var statusMessage string
+		if conf.Local.SaveAsJson {
+			filename = filepath.Join(conf.TmpDir, fmt.Sprintf("%d.json", e.ID))
+			defer os.Remove(filename)
+			writtenCount, err := WriteBundleToJson(fs, e.ID, filename)
+			if err != nil {
+				log.Printf("Failed to create tmp json file: %s", err)
+				return 0, err
+			}
+			statusMessage = fmt.Sprintf("Processing of bundle %d (%d bytes)", e.ID, writtenCount)
+		} else {
+			filename = filepath.Join(conf.TmpDir, fmt.Sprintf("%d.csv", e.ID))
+			outfile, err := os.Create(filename)
+			if err != nil {
+				log.Printf("Failed to create tmp csv file: %s", err)
+				return 0, err
+			}
+			defer os.Remove(filename)
+			defer outfile.Close()
+			csvOut := csv.NewWriter(outfile)
 
-		recordCount, err := WriteBundleToCSV(fs, e.ID, tableColumns, csvOut, wh)
-		if err != nil {
-			return 0, err
+			recordCount, err := WriteBundleToCSV(fs, e.ID, tableColumns, csvOut, wh)
+			if err != nil {
+				return 0, err
+			}
+			statusMessage = fmt.Sprintf("Processing of bundle %d (%d records)", e.ID, recordCount)
 		}
 
 		if err := LoadBundles(wh, filename, e); err != nil {
 			return 0, err
 		}
 
-		log.Printf("Processing of bundle %d (%d records) took %s", e.ID, recordCount,
-			time.Since(mark))
+		log.Printf("%s took %s", statusMessage, time.Since(mark))
 	}
 
 	// return how many files were processed
@@ -144,6 +156,9 @@ func ProcessFilesIndividually(wh warehouse.Warehouse, tableColumns []string, fs 
 func ProcessFilesByDay(wh warehouse.Warehouse, tableColumns []string, fs *fullstory.Client, exports []fullstory.ExportMeta) (int, error) {
 	if len(exports) == 0 {
 		return 0, nil
+	}
+	if conf.Local.SaveAsJson {
+		panic("The option to process files by day is only supported for CSV format.")
 	}
 
 	log.Printf("Creating group file starting with bundle %d (start: %s)", exports[0].ID, exports[0].Start.UTC())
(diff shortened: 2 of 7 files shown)


warehouse/localdisk.go
```diff
@@ -0,0 +1,130 @@
...
+
+// Writes the current time to timestamp file
+func (w *LocalDisk) SaveSyncPoints(bundles ...fullstory.ExportMeta) error {
+	t := time.Now().UTC().Format(time.RFC3339)
+	if _, err := os.Stat(w.conf.Local.SaveDir); os.IsNotExist(err) {
+		panic(err)
+	}
+	filename := filepath.Join(w.conf.Local.SaveDir, timestampFile)
+	timedata := []byte(t)
+	err := ioutil.WriteFile(filename, timedata, 0644)
```
butanian: return ioutil.WriteFile(filename, timedata, 0644) directly since the err isn't used in this scope

main.go
```diff
@@ -107,28 +108,41 @@ func ProcessExportsSince(wh warehouse.Warehouse, tableColumns []string, since ti
...
-		}
-		defer os.Remove(filename)
-		defer outfile.Close()
-		csvOut := csv.NewWriter(outfile)
+		var filename string
+		var statusMessage string
+		if conf.Local.SaveAsJson {
+			filename = filepath.Join(conf.TmpDir, fmt.Sprintf("%d.json", e.ID))
+			defer os.Remove(filename)
+			writtenCount, err := WriteBundleToJson(fs, e.ID, filename)
```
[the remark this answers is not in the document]
glarchev: It was mainly done that way because WriteBundleToCSV already existed, and WriteBundleToJson didn't, and it seemed like it would be more cumbersome to try and combine the two (they do both share the call to getExportData, and gzip reader, but that's about it). Currently the decision between the two only gets made in ProcessFilesIndividually (since it doesn't really make sense to group json files by day), but in the future it could also be made elsewhere (if needed).

warehouse/localdisk.go
```diff
@@ -0,0 +1,130 @@
...
+		t = w.conf.Local.StartTime
+	}
+	if _, err := os.Stat(w.conf.Local.SaveDir); os.IsNotExist(err) {
+		panic(err)
+	}
+	filename := filepath.Join(w.conf.Local.SaveDir, timestampFile)
+	if _, err := os.Stat(filename); os.IsNotExist(err) {
+		return t, nil
+	}
+	timebytes, err := ioutil.ReadFile(filename)
```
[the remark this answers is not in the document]
glarchev: I would argue that the original version is more readable (even though it has 1 extra line of code), but I'm open to making the change

warehouse/localdisk.go
glarchev: Will do

warehouse/localdisk.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	synctime, err := time.Parse(time.RFC3339, string(timebytes))
+	return synctime, err
+}
+
+// Writes the current time to timestamp file
+func (w *LocalDisk) SaveSyncPoints(bundles ...fullstory.ExportMeta) error {
+	t := time.Now().UTC().Format(time.RFC3339)
+	if _, err := os.Stat(w.conf.Local.SaveDir); os.IsNotExist(err) {
+		panic(err)
```
jameremo: A panic is usually reserved for programmer errors, not user input errors (I'm thinking of config params in the .toml file as user input). For a non-recoverable error like a bad `SaveDir`, our strategy thus far in `hauser` is to call `log.Fatalf`, which lets you write a message stdout and then exits. Hauser users are not always 100% savvy, so even though the existing panic will probably print sufficient info to tell that the dir is missing, the panic output may bewilder Go newbies. We should wrap it with a 100% user-friendly message.

Checking the `SaveDir` is a validation that could maybe go in `NewLocalDisk`, so we fail early, and don't have to repeat the check here and in `LastSyncPoint`.

warehouse/localdisk.go
```diff
@@ -0,0 +1,125 @@
...
+		fieldNames = append(fieldNames, name)
+	}
+	return fieldNames
+}
+
+// Reads the last sync date from file
+func (w *LocalDisk) LastSyncPoint() (time.Time, error) {
+	t := beginningOfTime
+	if w.conf.Local.UseStartTime {
+		t = w.conf.Local.StartTime
```
jameremo: It may be good to add a validation to `NewLocalDisk` to be sure that there is a `StartTime` specified when `UseStartTime` is true? I'm not sure how missing/empty strings would get parsed from the .toml file; I suspect that you would just have a zero value for the time.Time struct.

warehouse/localdisk.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+
+	return &LocalDisk{
+		conf: c,
+	}
+}
+
+// Currently returns the names of all the fields exported by BundleFields
+func (w *LocalDisk) GetExportTableColumns() []string {
```
jameremo: nit: [Go doc](https://blog.golang.org/godoc-documenting-go-code) for functions should always start with the function name. The same comment applies to other functions below, but I won't keep repeating it.

warehouse/localdisk.go
```diff
@@ -0,0 +1,125 @@
...
+		return "", err
+	}
+	defer srcFile.Close()
+	dstFile, err := os.Create(copiedFileName)
+	if err != nil {
+		return "", err
+	}
+	defer dstFile.Close()
+	_, err = io.Copy(dstFile, srcFile)
+	if err != nil {
```
jameremo: nit: these two lines could be combined into one, since we don't care about the number of bytes copied:
`if _, err := io.Copy(dstFile, srcFile); err != nil {`

main.go
```diff
@@ -143,6 +157,9 @@ func ProcessFilesByDay(wh warehouse.Warehouse, tableColumns []string, fs *fullst
 	if len(exports) == 0 {
 		return 0, nil
 	}
+	if conf.Local.SaveAsJson {
+		panic("The option to process files by day is only supported for CSV format.")
```
jameremo: See the comment on `SaveSyncPoints` for more details, but I think a `log.Fatalf` is preferable to a panic here, as it will give nicer output (panic prints a huge stack trace).

warehouse/localdisk.go
butanian: @jameremo do you have any strong feelings about including a more human readable message for something like this where we know what the exact problem is most of the time?

warehouse/localdisk.go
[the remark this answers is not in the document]
butanian: good to know!

main.go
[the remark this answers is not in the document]
butanian: sgtm also as Jess mentioned seems like there will be future work on Hauser, which hopefully will take care of this!

warehouse/localdisk.go
jameremo: @butanian Yes, that's exactly what I was trying to say above; if we know what the exact problm is (like a missing directory), we should clearly state it, so that the user doesn't have to think/work too hard (i.e. read code or decipher a go stack trace) to fix it.

warehouse/warehouse.go
```diff
@@ -18,3 +20,18 @@ type Warehouse interface {
 	GetUploadFailedMsg(filename string, err error) string
 	IsUploadOnly() bool
 }
+
+// Common implementation that subclasses can call
+func ValueToString(val interface{}, isTime bool) string {
```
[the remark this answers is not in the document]
glarchev: So right now each implementation 😃 just calls the method from `warehouse` (except for redshift, that looks sufficiently different). Is there a better way of doing this? I just hate how otherwise we'd have the same exact piece of code replicated in multiple implementations.

warehouse/localdisk.go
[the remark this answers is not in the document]
glarchev: Good to know, will make the change to the last line though.

warehouse/localdisk.go
```diff
@@ -0,0 +1,125 @@
...
+	"path/filepath"
+	"time"
+)
+
+type LocalDisk struct {
+	conf *config.Config
+}
+
+const (
+	timestampFile string = ".sync.hs"
```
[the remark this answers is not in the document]
glarchev: Yes, short for "hauser" -- I couldn't think of anything else, any suggestions?

config/config.go
```diff
@@ -65,6 +68,13 @@ type duration struct {
 	time.Duration
 }
 
+type LocalConfig struct {
+	SaveDir      string
+	StartTime    time.Time
+	UseStartTime bool
+	SaveAsJson   bool
```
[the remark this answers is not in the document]
glarchev: Ok, I guess in the future, once the Json uploads are supported, we should probably move the SaveAsJson flag to the main Config struct.

warehouse/localdisk.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return t, err
+	}
+	synctime, err := time.Parse(time.RFC3339, string(timebytes))
+	return synctime, err
+}
+
+// Writes the current time to timestamp file
+func (w *LocalDisk) SaveSyncPoints(bundles ...fullstory.ExportMeta) error {
+	t := time.Now().UTC().Format(time.RFC3339)
```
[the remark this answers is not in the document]
glarchev: Good to know!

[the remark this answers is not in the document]
glarchev: Can we assume that the timestamps in bundles are already in UTC? Or should we convert them?

warehouse/localdisk.go
glarchev: Will fix

warehouse/localdisk.go
glarchev: Will fix

warehouse/localdisk.go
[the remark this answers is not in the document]
jameremo: The FullStory client we're using ultimately creates the timestamps using the `time.Unix` function: https://github.com/fullstorydev/hauser/blob/87d1bc348c6cbf5e5d39db84cd1bb80bcdc86c2d/vendor/github.com/nishanths/fullstory/fullstory.go#L85

The [godoc](https://golang.org/pkg/time/#Unix) for that function says it returns a Local time, so I think you'll need to keep converting to UTC. Even if it were UTC, it might make sense to convert anyway, just on the off chance that the library we're using to fetch bundles changes its behavior. I haven't read all the code for `time.Time`, but my understanding is that the `Location` is for presentation only, so I don't think changing it changes the underlying representation, and thus doesn't incur any appreciable cost.

main.go
glarchev: Will fix

warehouse/localdisk.go
```diff
@@ -0,0 +1,130 @@
...
```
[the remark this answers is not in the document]
glarchev: It was pointed out that this should be the stop time of the last bundle, so I'm changing this.

warehouse/localdisk.go
glarchev: I think it depends on the particular toml parser implementation. I looked at the documentation for the one we're using, but didn't see anything specific to datetime. It's using reflection, so I'm guessing a missing datetime will be set to nil? I added a check to `NewLocalDisk` that will fail if `UseStartTime` is true and `StartTime` is nil.

glarchev: Actually, not nil, but the "zero" value of time

jameremo: I don't think it's possible for `StartTime` to be nil, as it's a `time.Time` struct and not a pointer to a struct (`*time.Time`). Welcome to golang! Read this article on [zero values](https://programming.guide/go/default-zero-value.html), hopefully it helps. I suspect that if the any value is missing from the toml, after parsing it will contain the zero value appropriate to its type. 

For the `time.Time` struct, its `wall` and `ext` values will be set to zero, which results in a zero value of `January 1, year 1, 00:00:00.000000000 UTC` (from the docs). Luckily it provides a handle `IsZero()` function, so you can check for a zero value, which likely means it hasn't been initialized. 

You should this out, as all of the above is just educated guessing.

jameremo: Sorry, I didn't reload the page in time to see your reply.

glarchev: Just tested it with StartTime not initialized, and it fails the IsZero check, as expected

glarchev:
@jameremo @butanian @sanjaybv I've made the requested changes, please let me know if I missed anything

warehouse/localdisk.go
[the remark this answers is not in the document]
jameremo: I don't think it would be bad to use `.hauser` in its entirety, and would be less cryptic. I think the `.hs` abbreviation is used for Haskell files.

warehouse/warehouse.go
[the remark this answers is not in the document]
jameremo: You've done it the way that it should be done; let me be more clear on the changes I was suggesting:

1. Change the comment to "Common logic that implementations can call" (i.e. remove the word "subclasses" from the comment.
2. Change the name of the function to `valueToString` so that it is not exported from the package, and thus is only callable via one of the exported implementations that calls it.

warehouse/localdisk.go
jameremo: @glarchev Could we change this `panic` call to `log.Fatalf`?

warehouse/warehouse.go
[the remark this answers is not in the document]
glarchev: Ah, makes sense! Will fix.

warehouse/localdisk.go
glarchev: Good catch, I actually meant to remove that check completely, since we're now checking whether the folder exists in the `NewLocalDisk` constructor.

glarchev: On the other hand, I think I'll keep it (and change it to `log.Fatalf`) -- since hauser is meant to run continuously, I suppose it's possible that someone would delete the folder while hauser is running (and `NewLocalDisk` constructor is only called once)

PR #46, opened 2019-01-10
add hauser image

jhump:


jameremo:
Haha, I do always think of Total Recall when I hear the name "Hauser". 🙂 I think as a nerdy sci-fi reference, the tone is fine, although maybe we're dating ourselves a bit, since there was a recent remake with Colin Farell? (but to be fair the Arnold version is the one I watched over and over as a kid, it was the first VHS we bought when we got our tape player, woohoo!)

Also, I seem to recall that the name was supposed to be a reference to a musician, I think Tim Hauser? I think this is near and dear to @jerseybbq's heart.

sanjaybv:
I have no idea what/who Hauser means/is. A quick google search didn't really answer my question. :(

Then, I tried looking for the meaning of it. Oh GAWD, none of the definitions in urban dictionary were helpful or said anything positive!

Searching "who is hauser" returns some musician named "Stjepan Hauser".

Or are we referring to the town in Idaho or the German surnames? Could someone please explain?

jhump:
@sanjaybv, I think @jerseybbq will have to answer as to what the name was originally intended to refer. But if you look at this PR's revised [README](https://github.com/fullstorydev/hauser/blob/jh/add-image/README.md), hopefully the image and caption provide adequate information about *my* interpretation of the name :)

PR #49, opened 2019-03-06
Reconcile and then update github.com/nishanths/hauser

butanian:
This PR undoes some of the changes that were manually added to nishanths/fullstory in the Hauser project, and updates to the latest version of that package. All the changes have been added to the library [here](https://github.com/nishanths/fullstory/pull/5/files). 

In addition, a [PR](https://github.com/nishanths/fullstory/pull/3) to make it possible to specify a custom user agent has been merged in.

The end goal is to be able to specify a custom user agent so we can gauge how many of our clients rely on Hauser to access their export data.


butanian:
I need sync up with jh tomorrow to understand his PR where we made Hauser compliant with modules! https://github.com/fullstorydev/hauser/commit/e8f85d157f550ac28674efa3afc0ee95183a89a7

Will ping after it is resolved!

butanian:
@jameremo resolved the build! PTAL

butanian:
+ @jhump in case I'm missing something. Followed the approach you shared, and it builds correctly now. To verify, I did make test with GOCACHE and GO111MODULE on and off. 

Manual edits to go.mod ->
[remove self reference](https://github.com/fullstorydev/hauser/pull/49/commits/77c2f8776d8097352c57e82156514d6ec857fc4a)
[undo module name change](https://github.com/fullstorydev/hauser/pull/49/commits/04bd4caa977ef34b8b6a72f6e5ded790bce0cd70)

vendor/github.com/nishanths/fullstory/client.go
```diff
@@ -57,7 +57,6 @@ func NewClient(apiToken string) *Client {
 // If the error is nil, the caller is responsible for closing the returned data.
 func (c *Client) doReq(req *http.Request) (io.ReadCloser, error) {
 	req.Header.Set("Authorization", "Basic " + c.APIToken)
-	req.Header.Set("Accept-encoding", "*")
```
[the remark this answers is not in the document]
butanian: Do you know who all worked on adding this header? I am still trying to figure out why it existed. Hauser does complain when a * value is not specified, despite the doc stating * is the default value if no header is specified. I guess we are forcing it to work even if some process is setting another header value?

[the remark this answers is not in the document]
butanian: actually nvm there is a very subtle different between * and no acceptable encoding (https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html). Will circle back after we resolve the base url in nishanths fullstory package.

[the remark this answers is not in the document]
butanian: added the accept encoding header back in, at least it is a cleaner approach now 😁

go.sum
```diff
@@ -1,9 +1,19 @@
+cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+cloud.google.com/go v0.31.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+cloud.google.com/go v0.34.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+cloud.google.com/go v0.36.0 h1:+aCSj7tOo2LODWVEuZDZeGCckdt6MlSF+X/rB3wUiS8=
+cloud.google.com/go v0.36.0/go.mod h1:RUoy9p/M4ge0HzT8L+SDZ8jg+Q6fth0CiBuhFJpSV40=
```
jhump: There are a lot of duplicates in here (fingerprints for multiple versions of the same module). I am not sure when/why the `go` tool chooses to do this. But, for what it's worth, if I remove the `go.sum` file and then re-generate it via `GO111MODULE=on go test ./...`, it results in a file with some of these extra version lines gone:
https://github.com/fullstorydev/hauser/compare/jh/skinnier-go-sum

go.mod
```diff
@@ -1,34 +1,10 @@
...
 require (
-	cloud.google.com/go v0.34.0
+	cloud.google.com/go v0.36.0
 	github.com/BurntSushi/toml v0.3.1
-	github.com/aws/aws-sdk-go v1.16.6
+	github.com/aws/aws-sdk-go v1.17.12
 	github.com/lib/pq v1.0.0
-	github.com/nishanths/fullstory v0.0.0-20180815131316-5c231e3c1db8 // this will need to be updated
-	github.com/pkg/errors v0.8.0
-	golang.org/x/sys v0.0.0-20181029174526-d69651ed3497
```
jhump: Did you mean to remove this line? I see this package referenced in the `go.sum` file, so it's a little strange that it's no longer relevant here.

vendor/github.com/nishanths/fullstory/client.go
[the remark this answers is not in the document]
jhump: According to the diff I am looking at, the `Accept-encoding` header was not added back in before this PR was merged. Also, what does upstream look like? Did you mean you added it back in Nishanth's source-of-truth repo?

vendor/github.com/nishanths/fullstory/go.mod
```diff
@@ -1 +0,0 @@
-module github.com/fullstorydev/hauser/vendor/github.com/nishanths/fullstory
```
jhump: You could have actually undone many of these. In my original PR, I had to add numerous `go.mod` files to make Go modules happy with the replacement directives I had in the root `go.mod` file. But now that there are no more replacement directives, none of them should be necessary.

vendor/gorepomanifest.json
```diff
@@ -31,8 +31,8 @@
  },
  "github.com/nishanths/fullstory": {
   "git-remote": "https://github.com/nishanths/fullstory",
-  "current-version-git-hash": "5c231e3c1db842ac29ef823e1eb75ecb04808614",
-  "content-hash": "90e212ed048fb0cdc0e362c874147a7365fed7cf"
+  "current-version-git-hash": "272735d360883d9fb41d542b464bc6cdc4f517ee",
```
jhump: Out of curiosity, is there a reason to vendor anything anymore? Do we need to pin some of these dependencies to *older* versions than their respect `master`?

I think it may be simplest at this point to just remove the `vendor` folder entirely...

go.mod
jhump: No I did not add it manually. I'm now curious what causes deps to pop in here or not. Maybe it was added before due to one of the `replacement` modules referring to the package. Now that we're not using replacements, perhaps it just uses whatever version of the package is specified by that other module's `go.mod`...

vendor/github.com/nishanths/fullstory/go.mod
jhump: I'm talking about just the `go.mod` files in the `vendor` folder. Running `go build` and `go test` won't do anything to generate those. (I had to manually create them.)

Note that some of the vendored folders *do* have a `go.mod` file already. But some did not, so I added one for those packages. Take a look at the original PR to see what all I had to add: #41.

vendor/gorepomanifest.json
jhump: 👍

PR #55, opened 2019-05-03
Added checks for valid SaveAsJson, GCSOnly, and S3Only config combinations

sam-fs:
Added checks for valid SaveAsJson, GCSOnly, and S3Only config combinations.  

If an invalid config combination is detected, instructions for correcting it are printed to the terminal.

### DIFF
--- a/main.go
+++ b/main.go
@@ -392,10 +392,17 @@ func main() {
 		wh = warehouse.NewLocalDisk(conf)
 	case "redshift":
 		wh = warehouse.NewRedshift(conf)
+		if conf.SaveAsJson {
+			if !conf.S3.S3Only {
+				log.Fatalf("Redshift doesn't support JSON format.  Ensure SaveAsJson = false in .toml file.")
+			}
+		}
 	case "bigquery":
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -392,8 +392,18 @@ func main() {
 		wh = warehouse.NewLocalDisk(conf)
 	case "redshift":
 		wh = warehouse.NewRedshift(conf)
+		if conf.SaveAsJson {
+			if !conf.S3.S3Only {
+				log.Fatalf("Redshift doesn't support JSON format.  Ensure SaveAsJson = false in .toml file.")
```
butanian: I see this - https://docs.aws.amazon.com/redshift/latest/dg/r_COPY_command_examples.html#r_COPY_command_examples-copy-from-json so the statement Redshift doesn't support JSON format may be confusing.

If it is feasible now (I will have to double check if we were in anyway limited by the CustomVars field or something else in our export) then maybe we should just say something along the lines that hauser doesn't support loading from JSON files yet?

Similar comment for BigQuery below.

PR #56, opened 2019-06-03
add pre-commit hook to format go files

camphillips22:
Added pre-commit hook to format go files during the commit.


camphillips22:
https://app.clubhouse.io/fullstory/story/71300/prevent-hauser-build-failures-due-to-formatting

camphillips22:
Not exactly. It does approximately the same thing, but the one in the fullstory codebase is much more sophisticated and has way too many features for what we are doing. This simply runs `gofmt` and `goimports` on the committed files and adds the changes to the commit.

PR #57, opened 2019-06-04
changed time precision to microseconds

camphillips22:
The time-stamps that are extracted from the data export bundles are currently truncated to seconds. This change extends the precision to microseconds.

### DIFF
--- a/warehouse/warehouse.go
+++ b/warehouse/warehouse.go
@@ -21,12 +21,16 @@ type Warehouse interface {
 	IsUploadOnly() bool
 }
 
+const (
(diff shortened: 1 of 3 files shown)


warehouse/warehouse.go
```diff
@@ -21,12 +21,16 @@ type Warehouse interface {
 	IsUploadOnly() bool
 }
 
+const (
+	RFC3339Micro = "2006-01-02T15:04:05.999999Z07:00"
```
butanian: could just be 
const RFC3339Micro = "2006-01-02T15:04:05.999999Z07:00"

PR #59, opened 2019-06-25
makefile improvements

camphillips22:
I was running into trouble with installing the external code checking tools (staticcheck, etc.) when running `make`. This PR employs a fix where the directory is changed to a temporary directory before `go get -u` is called in order to avoid using the local `go.mod`.

Another benefit of the changes is that the tools are only fetched if they don't exist in `GOBIN`.

I also changed the `dirs_to_check` to be explicit because I was also having trouble with that picking up unwanted directories.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,31 +14,7 @@ all_to_check=$(files_to_check) $(dirs_to_check)
 .PHONY: ci
 ci: deps checkgofmt vet staticcheck ineffassign predeclared test
 
-BIN_DIR := $(GOPATH)/bin
-STATICCHECK := $(BIN_DIR)/staticcheck
-INEFFASSIGN := $(BIN_DIR)/ineffassign
-PREDECLARED := $(BIN_DIR)/predeclared
-GOLINT := $(BIN_DIR)/golint
-ERRCHECK := $(BIN_DIR)/errcheck
-
-INSTALLTOOL = cd /tmp && GO111MODULE=off go get -u
-
-# Installs the "staticcheck" tool outside of the current GOPATH
-$(STATICCHECK):
-	$(INSTALLTOOL) honnef.co/go/tools/cmd/staticcheck
-
-$(INEFFASSIGN):
-	$(INSTALLTOOL) github.com/gordonklaus/ineffassign
-
-$(PREDECLARED):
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -2,7 +2,7 @@ dev_build_version=$(shell git describe --tags --always --dirty)
 # to_check is all code in this repo that we want to run checks on
 # (it is all Go code in here, but intentionally excludes the
 # vendor folder contents)
-dirs_to_check=$(shell find . -maxdepth 1 -mindepth 1 -type d | grep -v .git | grep -v vendor | grep -v .images)
+dirs_to_check=./warehouse ./config
```
jhump: This is slightly risky: if you add a new folder, it's easy to forget to update this line, and you then won't be getting any of these checks for the new stuff.

Makefile
```diff
@@ -14,6 +14,32 @@ all_to_check=$(files_to_check) $(dirs_to_check)
 .PHONY: ci
 ci: deps checkgofmt vet staticcheck ineffassign predeclared test
 
+BIN_DIR := $(GOPATH)/bin
```
jhump: I don't understand the need for the rest of the changes in this file.

The directory you are in should only matter if you do _not_ have the `GO111MODULE` env var set. (The way it decides whether to use modules or not, by default, is if you invoke it from a sub-directory of `GOPATH` IIRC.)

jhump:
> to avoid using the local go.mod

You only need to use `GO111MODULE=off` to avoid that. However, I must ask: why do you want to avoid using the local `go.mod` file? Do you mean that you were seeing unintended _changes_ to this file (like the `go` tool trying to add these tools to `go.mod`)?

> I was also having trouble with that picking up unwanted directories.

What unwanted directories? The build in CI always worked, and I don't remember having any problems myself. I'm wondering what is different about your environment.

camphillips22:
Ok, I did some rework to just use `GO111MODULE=off` for the tools. Also, I changed the `find | grep` on the `dirs_to_check` to fix it for my problem. Before, it was including `./.idea` so I thought it would be better to just be explicit, but it's probably best to not require changes to the makefile for any added folders.

PR #60, opened 2019-06-25
Reworked fetching and saving exports into a pipeline

camphillips22:
This PR is the first part of a re-architecture that is attempting to improve the extensibility of Hauser. The tasks of downloading, decoding, grouping, and saving the FullStory event exports are separated out into separate stages of a `Pipeline`. The output of the `Pipeline` is a filename and the corresponding bundle metadata. This is all that's needed to pass to the `Warehouse` in order upload to the database or bucket storage. 

While this PR doesn't make any significant changes to the `Hauser` feature set, it does enable flexibility in each of the processing steps.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -16,9 +16,11 @@ type Config struct {
 	TmpDir          string
 	ListExportLimit int
 	GroupFilesByDay bool
-	SaveAsJson      bool
-	PrettyJSON      bool
-	StartTime       time.Time
+
+	// SaveAsJSON can only be true if the final destination is not Redshift or BigQuery
+	SaveAsJson bool
+	// This will write the JSON data into a more human readable format with indentation
+	PrettyJSON bool
 
 	// for debug only; can point to localhost
 	ExportURL string
--- a/config/config.go
+++ b/config/config.go
@@ -19,8 +19,6 @@ type Config struct {
 
 	// SaveAsJSON can only be true if the final destination is not Redshift or BigQuery
 	SaveAsJson bool
-	// This will write the JSON data into a more human readable format with indentation
-	PrettyJSON bool
 
 	// for debug only; can point to localhost
 	ExportURL string
--- a/config/config.go
+++ b/config/config.go
@@ -17,6 +17,8 @@ type Config struct {
 	ListExportLimit int
 	GroupFilesByDay bool
 	SaveAsJson      bool
+	PrettyJSON      bool
+	StartTime       time.Time
 
 	// for debug only; can point to localhost
 	ExportURL string
--- a/main.go
+++ b/main.go
@@ -22,9 +22,6 @@ var (
 )
 
 const (
-	// Maximum number of times Hauser will attempt to retry each request made to FullStory
-	maxAttempts int = 3
-
 	// Default duration Hauser will wait before retrying a 429 or 5xx response. If Retry-After is specified, uses that instead. Default arbitrarily set to 10s.
 	defaultRetryAfterDuration = time.Duration(10) * time.Second
 )
--- a/main.go
+++ b/main.go
@@ -181,7 +181,7 @@ func main() {
 			case savedExport := <-savedExports:
 				log.Printf("Bundle saved to: %s", savedExport.Filename)
 
-				// Retry to upload the bundle to the warehouse with back off (and fail if not recoverable)
+				// Try to upload the bundle to the warehouse and retry with back off if it fails
 				for {
 					err := LoadBundles(wh, savedExport.Filename, savedExport.Meta...)
 					if BackoffOnError(err) {
--- a/main.go
+++ b/main.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"flag"
 	"log"
-	"net/http"
 	"os"
 	"strings"
 	"time"
@@ -19,11 +18,6 @@ var (
 	conf               *config.Config
 	currentBackoffStep = uint(0)
 	bundleFieldsMap    = warehouse.BundleFields()
-)
-
-const (
-	// Default duration Hauser will wait before retrying a 429 or 5xx response. If Retry-After is specified, uses that instead. Default arbitrarily set to 10s.
-	defaultRetryAfterDuration = time.Duration(10) * time.Second
 )
 
 // Record represents a single export row in the export file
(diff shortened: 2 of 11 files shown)


camphillips22:
As a comment, ignore the `Makefile` changes. I have them in a separate PR, but wanted them for my own development on this.

config/config.go
```diff
@@ -16,7 +16,11 @@ type Config struct {
 	TmpDir          string
 	ListExportLimit int
 	GroupFilesByDay bool
-	SaveAsJson      bool
+
+	// SaveAsJSON can only be true if the final destination is not Redshift or BigQuery
+	SaveAsJson bool
+	// This will write the JSON data into a more human readable format with indentation
+	PrettyJSON bool
```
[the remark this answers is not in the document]
camphillips22: It had more to do with reading it locally. I can remove it from this PR since it's a new feature.

main.go
```diff
@@ -418,23 +157,53 @@ func main() {
...
+		savedExports, errs := p.Start(startTime)
+
+		for {
+			select {
+			case savedExport := <-savedExports:
+				log.Printf("Bundle saved to: %s", savedExport.Filename)
+
+				// Try to upload the bundle to the warehouse and retry with back off if it fails
+				for {
+					err := LoadBundles(wh, savedExport.Filename, savedExport.Meta...)
```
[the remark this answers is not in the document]
camphillips22: It does assume that `fs.ExportList()` returns exports in the correct order (oldest to newest). We specify that that is the case in the [list export API endpoint](https://help.fullstory.com/develop-rest/data-export-api).

The pipeline handles things serially after that. Since the channels are unbuffered, they will move in lock step, preventing any ordering issues.

main.go
```diff
@@ -77,7 +61,7 @@ func TransformExportJSONRecord(wh warehouse.Warehouse, tableColumns []string, re
 			line = append(line, string(customVars))
 		} else {
 			if val, valExists := lowerRec[col]; valExists {
-				line = append(line, wh.ValueToString(val, field.IsTime))
+				line = append(line, convert(val, field.IsTime))
```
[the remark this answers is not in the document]
camphillips22: I had started down a path that I only partially came back from. Reverting this change.

pipeline/exports.go
```diff
@@ -0,0 +1,128 @@
...
+
+			fsErr = err
+			doRetry, retryAfterDuration := getRetryInfo(err)
+			if !doRetry {
+				return ExportData{}, err
+			}
+
+			log.Printf("Attempt #%d failed. Retrying after %s\n", r, retryAfterDuration)
+			time.Sleep(retryAfterDuration)
+			continue
```
butanian: you shouldn't need the continue

pipeline/pipeline.go
```diff
@@ -0,0 +1,182 @@
...
+	for {
+		fs := getFSClient(p.conf)
+		exportList, err := fs.ExportList(p.metaTime)
+		if err != nil {
+			p.errCh <- fmt.Errorf("could not fetch export list: %s", err)
+			return
+		}
+		for _, meta := range exportList {
+			p.metaCh <- meta
+			p.metaTime = meta.Stop
```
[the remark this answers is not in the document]
camphillips22: `metaTime` is used on line 67. It's how we avoid always calling `wh.LastSyncPoint`. Basically, if nothing bad happens in the pipeline or in loading bundles, we shouldn't have to ask the warehouse what the last checkpoint is. It should just be for restarting after a failure somewhere.

pipeline/pipeline.go
```diff
@@ -0,0 +1,182 @@
...
+	close(p.expCh)
+	close(p.metaCh)
+	close(p.recsCh)
+	close(p.saveCh)
+	close(p.errCh)
+}
+
+func (p *Pipeline) metaFetcher() {
+	for {
+		fs := getFSClient(p.conf)
```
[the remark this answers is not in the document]
camphillips22: Good catch. I agree that it should be outside the loop, but I don't think I want to share it between the two stages. I don't want to run into any issues with using the same client in two goroutines.

pipeline/exports_test.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"errors"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/nishanths/fullstory"
+)
+
+func TestGetRetryInfo(t *testing.T) {
```
jameremo: We need many many more tests than this, given the scale of this rewrite.

pipeline/pipeline.go
```diff
@@ -0,0 +1,182 @@
...
+		errCh:           make(chan error),
+	}
+}
+
+// Start begins pipeline downloading and processing bundles after the provided time. This function returns a channel
+// that will contain the filenames to which the data was saved and a channel that contains any errors that occur in the
+// pipeline. These must be retrieved from the channel to continue processing.
+func (p *Pipeline) Start(startTime time.Time) (chan SavedExport, chan error) {
+	p.metaTime = startTime
+	go p.metaFetcher()
```
jameremo: I think our goal is that if an error occurs at any stage of the pipeline, all stages shut down. However, if an error occurs in a later stage of the pipeline, this goroutine will never be shut down. The other goroutines are reading from a channel, so closing their upstream channel causes them to finish when they get a stop signal from a channel. This one isn't listening on a channel, it's just fetching exports, so it will only quit when there are no more exports available.

In this case you could have a second channel for communicating an early stop, or you could also use a context with a cancel function, which has some support for this bult-in: https://blog.golang.org/context.

pipeline/pipeline.go
```diff
@@ -0,0 +1,182 @@
...
+		if err != nil {
+			p.errCh <- fmt.Errorf("error fetching data export: %s", err)
+			return
+		}
+		p.expCh <- data
+	}
+}
+
+func (p *Pipeline) recordGrouper() {
+	var recs []Record
```
jameremo: It won't be workable to hold all the records in memory here. We're talking tens or hundreds of megabytes, or for extremely large clients a gigabyte or more of data in a single bundle. This has to work on a model that doesn't hold all the records in memory, hence the original design that just streamed data to disk.

pipeline/pipeline.go
```diff
@@ -0,0 +1,182 @@
...
+
+		if p.conf.GroupFilesByDay {
+			dataDay := data.meta.Start.UTC().Truncate(24 * time.Hour)
+			if groupDay.Before(dataDay) {
+				if len(recs) > 0 {
+					p.recsCh <- RecordGroup{
+						records: recs,
+						bundles: metas,
+					}
+					recs = recs[:0]
```
jameremo: When you do this with a slice, it is still pointing to the same underlying array, i.e. the same memory space. So below when we say `recs = append(recs, newRecs...)`, we end up overwriting the data passed into the channel as the value of `RecordGroup.records`. Demo: https://play.golang.org/p/NLOxZ1K_DIO

pipeline/pipeline.go
camphillips22: I had that at one point, but had removed it when I was cleaning things up and realized that the channels should close when they are using `range`. Clearly I forgot the first one 🤦‍♂

PR #61, opened 2019-07-26
add fullstory text logo file

sam-fs:
will use this png to replace quaid.png to make README more consistent with FS branding overall


butanian:
while I didn't have anything to do with quaid.png, I like the fullstory logo so LGTM 😄

PR #72, opened 2019-09-06
Temporarily remove module builds

butanian:
ASF mirror (git.apache.org) is experiencing [issues](https://status.apache.org/), and that is causing builds with modules enabled to fail.

This PR removes the two rows with module builds from the travis build matrix, so we can merge the [PR](https://github.com/fullstorydev/hauser/pull/69) which addresses some timezone issues with Redshift.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,8 @@ matrix:
       env:
       - GO111MODULE=off
       - VET=1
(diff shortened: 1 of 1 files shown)


butanian:
closing, trying to update all dependencies first

PR #78, opened 2019-10-16
Do not use tip for building CI

sanjaybv:


.travis.yml
```diff
@@ -15,7 +15,7 @@ matrix:
       env: GO111MODULE=off
     - go: "1.12"
       env: GO111MODULE=on
-    - go: tip
+    - go: "1.x"
```
butanian: trying to understand what this does. Does this force picking only a major release version? i.e. 1.12 vs 1.12.4.

go.mod
```diff
@@ -11,3 +11,5 @@ require (
 	github.com/pkg/errors v0.8.1
 	github.com/stretchr/testify v1.4.0 // indirect
 )
+
+go 1.13
```
butanian: can you expand on why this is required? will we have to keep changing this when we advance to 1.14 and beyond?

sanjaybv:
I do not think this PR is valid anymore.

PR #79, opened 2019-11-15
Allow S3 Only to use Sync Table

clettieri:
Our use case had us only needing to ingest directly to S3.  In order to take advantage of hauser's Sync table feature, we needed to update this function to call SaveSyncPoints() when S3Only is set in the config.

A quick little hack here but this allows anyone who wants to use ONLY S3 to take advantage of the SyncTable.  This update will still preserve the intended functionality for Redshift warehouse users.

Hope others can find this useful.

### DIFF
--- a/main.go
+++ b/main.go
@@ -210,6 +210,10 @@ func LoadBundles(wh warehouse.Warehouse, filename string, bundles ...fullstory.E
 	}
 
 	if wh.IsUploadOnly() {
+		if err := wh.SaveSyncPoints(bundles...); err != nil {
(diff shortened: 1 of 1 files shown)


patrick-fs:
Hey @clettieri, thanks for your PR! We'll take a look at it.

Your contribution is _much_ appreciated!

camphillips22:
Hey @clettieri, I realize this is a bit old, but I recently merged #89 which enables this flow by saving a "sync file" to the same S3 bucket. I'm going to go ahead and close this PR since it's no longer necessary.

PR #81, opened 2020-01-15
Use default HttpClient compression.

claygoddard:
The client will automatically set the accept encoding gzip header.
If the response is also gzip encoded, it will automatically be decoded
in the response body. This allows for clients behind proxies that already
gunzip the response body to work correctly.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


alexsazs:
Awesome! This is working like a charm now :)

PR #83, opened 2020-05-06
Hauser support additional http headers.

claygoddard:
Hauser support additional http headers for custom http client behavior.


client/client.go
```diff
@@ -30,24 +32,21 @@ func (e StatusError) Error() string {
...
-	APIToken string
-	BaseURL  string
+	Config     *config.Config
+	BaseURL    string
 }
 
 // NewClient returns a Client initialized with http.DefaultClient and the
 // supplied apiToken.
-func NewClient(apiToken string) *Client {
+func NewClient(config *config.Config) *Client {
```
claygoddard, line 41: Just a strict move from an external repo. The only change is referencing the config directly rather than an intermediate object.

client/export.go
```diff
@@ -10,62 +10,6 @@ import (
 	"time"
 )
 
-// Session represents the FullStory session for a user.
-type Session struct {
```
claygoddard, line 14: Removed the stuff from the dep that isn't used.

client/client.go
```diff
@@ -30,24 +32,21 @@ func (e StatusError) Error() string {
...
+		baseURL = config.ExportURL
+	}
 	return &Client{
 		HTTPClient: http.DefaultClient,
-		Config: Config{
-			APIToken: apiToken,
-			BaseURL:  BaseURL,
-		},
+		Config:     config,
+		BaseURL:    baseURL,
```
camphillips22, line 49: I feel like we shouldn't have both here. If the config already contains a URL, we should use that, and have the default populate when loading the config.

claygoddard: Yep, that's way better now that the client code has been moved into the rep. Will update to that.

PR #86, opened 2020-09-01
Convert hauser main process to service and add testing

camphillips22:
This change reworks the main hauser process to use dependency injection so
that the client interfaces can be mocked. I added some very basic
testing to prove that the standard process is working correctly.

I plan on following up with some more structural changes to make
the configuration behavior a little more consistent.


client/client.go
```diff
@@ -26,12 +26,21 @@ func (e StatusError) Error() string {
...
 
+// DataExportClient represents an interface for interacting with the FullStory Data Export API
+type DataExportClient interface {
+	ExportList(start time.Time) ([]ExportMeta, error)
+	ExportData(id int, modifyReq ...func(r *http.Request)) (ExportData, error)
+}
+
 // Client represents a HTTP client for making requests to the FullStory API.
 type Client struct {
+	DataExportClient
```
claygoddard, line 37: I don't think you want this inlined here. The `Client` itself is the one that should be implementing the interface. The type check on line 42 has you covered. Plus, calling `Client.DataExportClient.ExportList()`, for example, would panic..

client/client.go
```diff
@@ -26,12 +26,21 @@ func (e StatusError) Error() string {
 	return fmt.Sprintf("fullstory: response error: Status:%s, StatusCode:%d, RetryAfter:%v", e.Status, e.StatusCode, e.RetryAfter)
 }
 
+// DataExportClient represents an interface for interacting with the FullStory Data Export API
+type DataExportClient interface {
+	ExportList(start time.Time) ([]ExportMeta, error)
```
claygoddard, line 31: Quick method comments for the new interface?

client/client.go
camphillips22: Good call. I meant to remove this.

config/config.go
```diff
@@ -107,5 +108,12 @@ func Load(filename string) (*Config, error) {
 		conf.ExportURL = DefaultExportURL
 	}
 
+	if conf.TmpDir == "" {
+		conf.TmpDir, err = ioutil.TempDir("", "hauser")
```
claygoddard, line 112: Is this a safe change? What was happening for empty string folks before? Storing in the relative current directory?

main.go
```diff
@@ -37,7 +35,20 @@ const (
...
 
-type ExportProcessor func(warehouse.Warehouse, []string, *client.Client, []client.ExportMeta) (int, error)
+type HauserService struct {
+	config       *config.Config
+	fsClient     client.DataExportClient
+	warehouse    warehouse.Warehouse
+	tableColumns []string
+}
+
+func NewHauser(config *config.Config, fsClient client.DataExportClient, warehouse warehouse.Warehouse) *HauserService {
```
claygoddard, line 45: While we are refactoring here, should we move `HauserService` out of `main.go` into its own file or package?

config/config.go
camphillips22: Ah, it looks like it does just resolve to the current directory (makes sense 😄 ). I can remove this change.

main.go
```diff
@@ -349,20 +368,55 @@ func BackoffOnError(err error) bool {
...
+	return h.ProcessExportsSince(lastSyncedRecord)
+}
+
+func (h *HauserService) Run() {
+	if err := h.Init(); err != nil {
+		log.Fatal(err)
+	}
+	for {
+		if numBundles, err := h.ProcessNext(); err != nil {
+			if h.BackoffOnError(err) {
```
claygoddard, line 399: I think you want to unconditionally call `BackoffOnError` here so that it resets when there is no error.

main_test.go
```diff
@@ -1,17 +1,132 @@
...
 	"github.com/fullstorydev/hauser/client"
+	"github.com/fullstorydev/hauser/config"
+	hausertest "github.com/fullstorydev/hauser/testing"
 	"github.com/fullstorydev/hauser/warehouse"
 	"github.com/pkg/errors"
 )
 
+var update = flag.Bool("update", false, "update upload files")
+
+func Ok(t *testing.T, err error, format string, a ...interface{}) {
```
claygoddard, line 25: Should these go in some testing utils package?

claygoddard:
A few comments, once addressed things should be good to go!

main.go
camphillips22: Ah, yes. Of course!

main.go
camphillips22: I can do that. Do you think we should move it to an `internal` package? Or just something like "core"?

claygoddard: `internal/service.go` seems reasonable to me!

PR #87, opened 2020-09-01
Enable BiqQuery partition expiration

camphillips22:
Adds a configuration field for setting the expiration time for
the table partitions.

### DIFF
--- a/example-config.toml
+++ b/example-config.toml
@@ -42,6 +42,7 @@ Project = "<your project>"
 Dataset = "<your dataset>"
 ExportTable = "fs_export"
 SyncTable = "fs_sync"
+PartitionExpiration = "0"
 
 [local]
 SaveDir = "<Path to your local folder to save files to>"
--- a/example-config.toml
+++ b/example-config.toml
@@ -42,6 +42,10 @@ Project = "<your project>"
 Dataset = "<your dataset>"
 ExportTable = "fs_export"
 SyncTable = "fs_sync"
+# The amount of time after which the partitions will expire
+# Valid time units are "s", "m", "h" (seconds, minutes, hours).
+# For example, "720h" would expire the partitions after 30 days.
+# If this value is omitted or "0", then the partitions will not expire.
 PartitionExpiration = "0"
 
 [local]
--- a/warehouse/bigquery.go
+++ b/warehouse/bigquery.go
@@ -12,7 +12,6 @@ import (
 
 	"cloud.google.com/go/bigquery"
 	"cloud.google.com/go/storage"
-
 	"github.com/fullstorydev/hauser/client"
 	"github.com/fullstorydev/hauser/config"
 )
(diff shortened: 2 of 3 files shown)


warehouse/bigquery.go
```diff
@@ -12,7 +12,6 @@ import (
 
 	"cloud.google.com/go/bigquery"
 	"cloud.google.com/go/storage"
-
```
claygoddard, line 15: nit: Put newline back.

warehouse/bigquery.go
```diff
@@ -206,26 +205,29 @@ func (bq *BigQuery) EnsureCompatibleExportTable() error {
...
-		if err := bq.updateTable(table, md.Schema); err != nil {
-			return nil
+		update.Schema = append(md.Schema, missingFields...)
+		needsUpdate = true
+	}
+
+	if md.TimePartitioning.Expiration != bq.conf.BigQuery.PartitionExpiration.Duration {
+		update.TimePartitioning = &bigquery.TimePartitioning{
+			Expiration: bq.conf.BigQuery.PartitionExpiration.Duration,
+			Field:      md.TimePartitioning.Field,
```
claygoddard, line 222: Now should the field be set based on the export data? Like by using `SessionStart`?

example-config.toml
```diff
@@ -42,6 +42,7 @@ Project = "<your project>"
 Dataset = "<your dataset>"
 ExportTable = "fs_export"
 SyncTable = "fs_sync"
+PartitionExpiration = "0"
```
jameremo, line 45: Just making sure this is backward-compatible with older hauser configs: when this line is omitted, the config struct gets default values, so the expiration duration is 0. And zero means don't expire, correct?

Also, in order for this to parse correctly, this is the usual go shorthand for durations, correct? Might be nice to put a comment here to explain or link to that syntax, since so many hauser users are not overly familiar with go.

warehouse/bigquery.go
jameremo: When hauser was created, you couldn't partition a table in BQ using a field value, it was always partitioned by ingestion time. I think that's how the tables are still being created by hauser today. So in `LoadToWarehouse` we are doing some super hacky stuff with the legacy notation to load data into a particular partition. We had a card in at some point to change to creating the table such that it partitions by the `EventStart` field, which is semantically what our existing hack does now.

I think carrying over the partitioning field from the loaded table metadata here sets us up for a smooth transition to however we might change partitioning in the future.

camphillips22: > I think carrying over the partitioning field from the loaded table metadata here sets us up for a smooth transition to however we might change partitioning in the future.

Yeah @jameremo that's what I was thinking. I was intending on following up with that change.

example-config.toml
camphillips22: > Just making sure this is backward-compatible with older hauser configs: when this line is omitted, the config struct gets default values, so the expiration duration is 0. And zero means don't expire, correct?

This is correct. I tested with a real table to make sure.

> Also, in order for this to parse correctly, this is the usual go shorthand for durations, correct? Might be nice to put a comment here to explain or link to that syntax, since so many hauser users are not overly familiar with go.

Good idea, will update!

warehouse/bigquery.go
camphillips22: I think that `goimports` did that...

warehouse/bigquery.go
claygoddard: Gotcha. Yeah, Cam shared a card with me directly. SGTM

PR #88, opened 2020-09-01
DNM: For cam's info

iansmith:
This shows you the plan I want to use.  I want to abstract Client to FSClient so I can supply my own which has different behavior in terms of authorization.

However, I need to do it this way because I want to *not* fork the code so I'm going to supply a different main in a project called "fshauser" or similar.  Just like with grpcurl and fsgrpcurl.


camphillips22:
Closing since this was addressed with #92 and #91 .

PR #90, opened 2020-09-03
Add goreleaser

camphillips22:
Add goreleaser to enable building binaries that can be uploaded
to github.

I also added a version flag to `main.go` to be able to print out the release version.


Makefile
```diff
@@ -15,7 +15,7 @@ all_to_check=$(files_to_check) $(dirs_to_check)
 .PHONY: ci
 ci: deps checkgofmt vet staticcheck ineffassign predeclared test
 
-INSTALLTOOL = GO111MODULE=off go get
+INSTALLTOOL = GO111MODULE=on go get
```
camphillips22, line 18: Given our conversation earlier, I thought it was also okay to go ahead and flip this to on. Just calling `make` was failing for me (include `make release`), which prompted me to change it.

Makefile
```diff
@@ -29,6 +29,11 @@ updatedeps:
 install:
 	go install -ldflags '-X "main.version=dev build $(dev_build_version)"' .
 
+.PHONY: release
+release:
+	$(INSTALLTOOL) get github.com/goreleaser/goreleaser
```
camphillips22, line 34: The first time I ran this it introduced a bunch of go.mod changes. However, in grpcurl, I saw that any go.mod/go.sum changes are reverted after running `make release`, so I reverted them too. 

I suppose I could add a helper release script like the one in grpcurl, but since I'm just starting with binaries and not docker or homebrew, I thought I would keep it simple for now.

Makefile
jhump: You actually want `go install` with modules. A `go get` with modules will actually update to the latest version. (I had to recently change this in the grpcurl and grpcui `Makefile`s too.)

Makefile
jhump: See above: it's because you were doing a `go get` instead of install. BTW, there appears to be a bug in this line: it has a second `get` which I think would make the command fail.

main.go
```diff
@@ -2,18 +2,29 @@ package main
...
+	"os"
+	"path/filepath"
 
 	"github.com/fullstorydev/hauser/client"
 	"github.com/fullstorydev/hauser/config"
 	"github.com/fullstorydev/hauser/internal"
 	"github.com/fullstorydev/hauser/warehouse"
 )
 
+var version = "dev build <no version set>"
```
jhump, line 16: Doh! I guess this means `make install` has never worked?

Sorry. That `Makefile` was literally copied from one of the other repos, so I must not have done enough to verify it...

jhump (approved):
I think you need to fix the `go get` vs `go install` thing and remove an errant "get" from the release target.

But otherwise LGTM.

Makefile
camphillips22: 🤦

Makefile
camphillips22: Ah, good to know!

main.go
camphillips22: It's all good!

PR #91, opened 2020-09-17
factor out all resources for main/init

iansmith:
@camphillips22 @claygoddard 

This is nothing but a reorg to make it easy to build a main without much duplicated code.  I moved the "make" function to a package called resources and created a "MakeClient" for having the network separated out.


config/config.go
```diff
@@ -180,3 +182,9 @@ func Validate(conf *Config) error {
 	}
 	return nil
 }
+var version = "dev build <no version set>"
+
+func PrintVersion() {
+	fmt.Printf("%s %s\n", filepath.Base(os.Args[0]), version)
+	os.Exit(0)
+}
```
camphillips22, line 190: I don't think this belongs here. This is specifically a "main" concern so it should stay there. This was added to enable printing versions of the "hauser" binary once we start building them.

go.mod
```diff
@@ -6,11 +6,10 @@ require (
 	cloud.google.com/go/pubsub v1.0.1 // indirect
 	github.com/BurntSushi/toml v0.3.1
 	github.com/aws/aws-sdk-go v1.23.18
-	github.com/gordonklaus/ineffassign v0.0.0-20200809085317-e36bfde3bb78 // indirect
```
camphillips22, line 9: I don't think we want these changes to go.mod and go.sum since you didn't change any dependencies

client/export.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/http/httptrace"
```
camphillips22, line 8: Unused import

client/client.go
```diff
@@ -40,29 +40,21 @@ type DataExportClient interface {
...
 
 var _ DataExportClient = (*Client)(nil)
 
-// NewClient returns a Client initialized with http.DefaultClient and the
-// supplied apiToken.
-func NewClient(config *config.Config) *Client {
-	return &Client{
-		HTTPClient: http.DefaultClient,
-		Config:     config,
-	}
```
camphillips22, line 53: I think this should stay here. Since you are trying to create a custom client, why don't you do something like:

```golang
myClient := client.NewClient(conf)
myClient.SetCustomAuthHeaders = func(req *http.Request) {
  // do custom auth
}
```

And then the client would have:

```golang
// no need to export
func (c *Client) setAuthHeaders(req *http.Request) {
  if c.SetCustomAuthHeaders != nil {
    c.SetCustomAuthHeaders(req)
    return
  }
  req.Header.Set("Authorization", "Basic "+c.Config.FsApiToken)
  for _, header := range c.Config.AdditionalHttpHeader {	
    req.Header.Set(header.Key, header.Value)	
  }
}.

File: config/config.go (hunk 1 above)
Comment (iansmith) [reply config/config.go:190]: I was try to deduplicate the code, since other instances of "main" might want to print version of code also.

I moved it back to main.  Fixed.

File: client/export.go (hunk 3 above)
Comment (iansmith) [reply client/export.go:8]: Sorry, expected the commit hook in git to find all those.

Fixed.

File: client/client.go (hunk 4 above)
Comment (iansmith) [reply client/client.go:53]: That may not cover the case where you want to actually change the client itself.  In client now, HTTPClient points to http.DefaultClient but in more complex auth schemes it does not.  For example, I use oauth2.Client which is also an *http.Client but has to be initialized differently.

If you are ok, with actually reaching into the Client object and mucking with the HTTPClient  field (currently it's uppercase) I am happy to do that and move the NewClient() back to client package.

Do you not think it's more "parallel" to the way we handle the Warehouse and Storage to have Client be a resource?

Comment (camphillips22) [reply client/client.go:53]: No, since all the client has to do is match the `DataExportClient` interface. In truth, none of the fields in the `Client` struct should be exported.

It's actually better if you just implement your own client that implements the `DataExportClient` interface. You can pass that directly to the `NewHauser` function.

File: client/export.go (hunk 3 above)
Comment (jhump) [reply client/export.go:8]: The git hook we typically use only works in our closed-source repo (since it refers to scripts in that repo.) You'd have to manually setup a commit hook for these repos or configure your own global hook in `$HOME/.gitconfig`.

File: client/client.go [hunk 5]
```diff
@@ -40,29 +40,21 @@ type DataExportClient interface {
...
+func (c *Client) AddAuthHeaders(req *http.Request) {
```
jhump, line 48: nit: I would not expose anything like this.

Just allow caller to supply a custom HTTP client (not concrete type but an interface -- like `DataExportClient` or maybe just `http.RoundTripper`) and then have hauser's `main` create a client that automatically adds the configured API key as an auth header. That way, any sort of customization is just done via decorating the HTTP client.

client/client.go
```diff
@@ -40,29 +40,21 @@ type DataExportClient interface {
...
+	c.AuthFunc(req)
 }
 
 // doReq performs the supplied HTTP request and returns the data in the response.
 // Necessary authentication headers are added before performing the request.
 //
 // If the error is nil, the caller is responsible for closing the returned data.
 func (c *Client) doReq(req *http.Request) (io.ReadCloser, error) {
-	req.Header.Set("Authorization", "Basic "+c.Config.FsApiToken)
```
jhump, line 61: 1) This code was removed, not moved. So this totally breaks this repo, right?
2) It could just move into a wrapper/decorator for `HTTPClient`, which then allows you to never need to call any auth func.

client/export.go
camphillips22: There's also a commit hook you can add for this repo that I added with #56

client/client.go
iansmith: When I did the reorg that is about to be posted here, I put this in the proper place (the auth func).  Please advise if this is sufficient.

client/client.go
iansmith: I don't know if it will suffice for what you wanted but:

* Made this function private (addAuthHeaders)
* Allowed the auth func to be passed into the constructor of Client()
* Made all the fields of Client private
* Put NewClient() back to client package and removed additional parameters from the default path used by Hauser's main.  Thus Hauser's main is now unchanged and there is the _option_ to supply an auth func, that Hauser's main does not use.

This minimizes duplication and the only thing is exposed is the option to supply your own auth func at construction time. Everything else is now private.

@camphillips22  ^^^

client/client.go
iansmith: See comment below about the new version.

If you would prefer (I asked in chat, but haven't heard back yet) I can duplicate the code from client.go and go the route of *just* being a different impl of `DataExportClient`.

client/client.go
```diff
@@ -38,32 +38,42 @@ type DataExportClient interface {
 
 // Client represents a HTTP client for making requests to the FullStory API.
 type Client struct {
-	HTTPClient *http.Client
-	Config     *config.Config
+	httpClient    *http.Client
+	configuration *config.Config
+	authFunc      func(*http.Request,*config.Config)
```
jhump, line 43: I think you can delete everything related to auth. After #92, the only necessary change is making `NewHauser` exported (e.g. moving it out of internal package).

client/client.go
jhump: I think you should just remove `authFunc` entirely. The auth details can be left to the transport. See #92.

main.go
```diff
@@ -4,54 +4,16 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"github.com/fullstorydev/hauser/client"
```
jhump, line 7: These look wrong. Try `$FS_HOME/bin/goimports -local github.com/fullstorydev/hauser/ -w .`

client/client.go
```diff
@@ -38,32 +38,47 @@ type DataExportClient interface {
...
+	return http.DefaultClient.Transport.RoundTrip(req)
+}
+func (f *FullStoryAPIKeyRoundTripper) addAuthHeaders(req *http.Request) {
+	req.Header.Set("Authorization", "Basic "+f.cfg.FsApiToken)
+	for _, header:=range f.cfg.AdditionalHttpHeader{
+		req.Header.Add(header.Key, header.Value)
+	}
+}
+
+func NewClient(cfg *config.Config) *Client {
```
camphillips22, line 64: I don't think you need any changes to this file if you rebase off of master.

warehouse/redshift_test.go
```diff
@@ -146,10 +146,10 @@ func TestValidateSchemaConfig(t *testing.T) {
 		wh := NewRedshift(tc.conf)
 		err := wh.validateSchemaConfig()
 		if tc.hasError && err == nil {
-			t.Errorf("expected Redshift.validateSchemaConfig() to return an error when config.Config.Redshift.DatabaseSchema is empty")
+			t.Errorf("expected Redshift.validateSchemaConfig() to return an error when config.configuration.Redshift.DatabaseSchema is empty")
```
jhump, line 149: Were changes in this file intentional?

resources/resources.go
```diff
@@ -0,0 +1,51 @@
+package resources
+
+import (
+	"cloud.google.com/go/storage"
+	"context"
+	"github.com/fullstorydev/hauser/client"
```
jhump, line 6: ditto

client/client.go
```diff
@@ -38,32 +38,47 @@ type DataExportClient interface {
 
 // Client represents a HTTP client for making requests to the FullStory API.
 type Client struct {
-	HTTPClient *http.Client
-	Config     *config.Config
+	HTTPClient    *http.Client
+	Configuration *config.Config
```
jhump, line 42: Why the rename here?

client/client.go
```diff
@@ -38,8 +38,8 @@ type DataExportClient interface {
 
 // Client represents a HTTP client for making requests to the FullStory API.
 type Client struct {
-	HTTPClient *http.Client
-	Config     *config.Config
+	HTTPClient    *http.Client
+	Config *config.Config
```
camphillips22, line 42: These changes are unnecessary.

resources/resources.go
```diff
@@ -0,0 +1,51 @@
+package resources
```
camphillips22, line 1: nit: Could we rename this package to `core`? I could imagine other things possibly belonging here that could be generally shared, but relate more to the main "hauser" process.

Ideally, I think these would actually be in a root "hauser" package and the main package would actually be located in `cmd/hauser`. Alas...

go.mod
iansmith: I reverted them.

client/client.go
iansmith: Yep, did that.

client/client.go
iansmith: Yep... great minds...

client/client.go
iansmith: That was reverted when I accepted cam's changes in #92

warehouse/redshift_test.go
iansmith: they have been reverted.  The change was actually because I changed the variable config to be private in an earlier version.

hauser/resources.go
```diff
@@ -0,0 +1,51 @@
+package hauser
+
+import (
+	"cloud.google.com/go/storage"
+	"context"
+	"github.com/fullstorydev/hauser/client"
+	"github.com/fullstorydev/hauser/config"
+	"github.com/fullstorydev/hauser/internal"
+	"github.com/fullstorydev/hauser/warehouse"
```
camphillips22, line 9: Import ordering here is wrong.

client/client.go
camphillips22: Can you remove these changes? This will contribute to `checkgofmt` errors.

hauser/resources.go
```diff
@@ -0,0 +1,51 @@
+package hauser
```
camphillips22, line 1: can you change this file to `core/core.go` and change the package to `core`?

resources/resources.go
iansmith: Done.

client/client.go
iansmith: Believe that gofmt was needed to get rid of the spaces there.

hauser/resources.go
iansmith: fixed.

hauser/resources.go
iansmith: done.

main.go
iansmith: I fixed the grouping if that's what looks wrong.  Given that you used the -local option, I'm assuming that was the problem.

warehouse/redshift_test.go
camphillips22: You still need to remove these changes.

iansmith: done. 
```
$ git diff master -- warehouse/*.go
$
```

resources/resources.go
iansmith: fixed by hand

.gitignore
```diff
@@ -1,3 +1,3 @@
 dist/
-hauser
```
camphillips22, line 2: Remove this. It was originally there to ignore the binary that gets built.

camphillips22 (approved):
Just one thing, otherwise LGTM!

PR #92, opened 2020-09-22
Apply auth headers in round tripper

camphillips22:
Moves the API token and additional headers into
a round tripper implementation to remove it from
the `doReq` function.

### DIFF
--- a/client/transport.go
+++ b/client/transport.go
@@ -6,9 +6,9 @@ import (
 	"github.com/fullstorydev/hauser/config"
 )
 
-// FullStory is an HTTP transport which wraps the underlying transport and
+// APIKeyRoundTripper is an HTTP transport which wraps the underlying transport and
 // sets the Authorization header
-type FullStory struct {
+type APIKeyRoundTripper struct {
 	Key               string
 	AdditionalHeaders []config.Header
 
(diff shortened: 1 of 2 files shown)


camphillips22:
Thanks @jhump for the suggestion to do this auth in a `RoundTripper`!

client/transport.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"net/http"
+
+	"github.com/fullstorydev/hauser/config"
+)
+
+// FullStory is an HTTP transport which wraps the underlying transport and
+// sets the Authorization header
+type FullStory struct {
```
jhump, line 11: Maybe name this more after what it does -- like `APIKeyRoundTripper`?

client/transport.go
```diff
@@ -0,0 +1,31 @@
...
+	// if nil, http.DefaultTransport is used.
+	Transport http.RoundTripper
+}
+
+func (t *FullStory) RoundTrip(req *http.Request) (*http.Response, error) {
+	rt := t.Transport
+	if rt == nil {
+		rt = http.DefaultTransport
+	}
+	newReq := *req
```
jhump, line 25: There's not much point in copying the request this way. The only thing you are touching is in headers, and `newReq` and `req` still share the same headers instance since you aren't doing a deep copy. And, FWIW, I don't think you need to do a deep copy. I think you could just add the headers directly to `req`.

jhump (approved):
A couple of minor comments.

PR #93, opened 2020-09-22
Removed warehouse dependency on ExportMeta

camphillips22:
To prep for using the new APIs, we need to remove the dependency on
`ExportMeta` from the warehouse package and instead rely directly
on the "stop time" (i.e. the time for the last bundle we loaded into
the warehouse). This change should be fairly transparent with the
exception of the ID that was inserted into the sync table.

### DIFF
--- a/internal/service.go
+++ b/internal/service.go
@@ -233,9 +233,9 @@ func (h *HauserService) LoadBundles(ctx context.Context, filename string, bundle
 		return fmt.Errorf("failed to save file: %s", err)
 	}
 
+	bundleEndTime := bundles[len(bundles)-1].Stop
 	if h.config.StorageOnly {
-
-		return h.storage.SaveSyncPoint(ctx, bundles[len(bundles)-1].Stop)
+		return h.storage.SaveSyncPoint(ctx, bundleEndTime)
 	}
 
 	defer h.storage.DeleteFile(ctx, objName)
@@ -250,7 +250,7 @@ func (h *HauserService) LoadBundles(ctx context.Context, filename string, bundle
 	// still okay - the next call to LastSyncPoint() will see that there are export
 	// records beyond the sync point and remove them - ie, we will reprocess the
 	// current export file
-	if err := h.database.SaveSyncPoint(ctx, bundles[len(bundles)-1].Stop); err != nil {
+	if err := h.database.SaveSyncPoint(ctx, bundleEndTime); err != nil {
 		log.Printf("Failed to save sync points for bundles ending with %d: %s", bundles[len(bundles)].ID, err)
 		return err
 	}
--- a/internal/service.go
+++ b/internal/service.go
@@ -234,12 +234,14 @@ func (h *HauserService) LoadBundles(ctx context.Context, filename string, bundle
 	}
 
 	if h.config.StorageOnly {
-		return h.storage.SaveSyncPoints(ctx, bundles...)
+
+		return h.storage.SaveSyncPoint(ctx, bundles[len(bundles)-1].Stop)
 	}
 
 	defer h.storage.DeleteFile(ctx, objName)
 
-	if err := h.database.LoadToWarehouse(objRef, bundles...); err != nil {
+	startTime := bundles[0].Start.UTC()
+	if err := h.database.LoadToWarehouse(objRef, startTime); err != nil {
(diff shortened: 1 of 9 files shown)


internal/service.go
```diff
@@ -234,12 +234,14 @@ func (h *HauserService) LoadBundles(ctx context.Context, filename string, bundle
 	}
 
 	if h.config.StorageOnly {
-		return h.storage.SaveSyncPoints(ctx, bundles...)
+
```
claygoddard, line 237: Intentional new line addition?

internal/service.go
```diff
@@ -248,7 +250,7 @@ func (h *HauserService) LoadBundles(ctx context.Context, filename string, bundle
 	// still okay - the next call to LastSyncPoint() will see that there are export
 	// records beyond the sync point and remove them - ie, we will reprocess the
 	// current export file
-	if err := h.database.SaveSyncPoints(ctx, bundles...); err != nil {
+	if err := h.database.SaveSyncPoint(ctx, bundles[len(bundles)-1].Stop); err != nil {
```
claygoddard, line 253: nit: pull out to `syncPoint` var to share above?

warehouse/bigquery.go
```diff
@@ -102,19 +101,16 @@ func (bq *BigQuery) LastSyncPoint(_ context.Context) (time.Time, error) {
...
 	if err := bq.connectToBQ(); err != nil {
 		return err
 	}
 	defer bq.bqClient.Close()
 
-	values := make([]string, len(bundles))
-	for i, p := range bundles {
-		values[i] = fmt.Sprintf("(%d, TIMESTAMP(\"%s\"), TIMESTAMP(\"%s\"))", p.ID, time.Now().UTC().Format(time.RFC3339), p.Stop.UTC().Format(time.RFC3339))
-	}
+	value := fmt.Sprintf("(%d, TIMESTAMP(\"%s\"), TIMESTAMP(\"%s\"))", -1, time.Now().UTC().Format(time.RFC3339), endTime.UTC().Format(time.RFC3339))
```
claygoddard, line 110: What is -1 going to give you for ID? Is that going to be problematic?

claygoddard: Also only inserting one record is not going to be an issue with the `removeSyncPointsAfter` behavior above? I guess the `bundles` are always day aligned so I guess it should be okay?

claygoddard (commented):
A couple questions before approving.

warehouse/bigquery.go
camphillips22: > What is -1 going to give you for ID? Is that going to be problematic?

We don't use the IDs that are stored in the database table. We just query for the max timestamp in the last column (bundle end time). I thought about synthesizing the IDs, but I don't think that really buys us anything in with the new API paradigm.

> Also only inserting one record is not going to be an issue with the removeSyncPointsAfter behavior above? I guess the bundles are always day aligned so I guess it should be okay?

I don't think this should be an issue since `removeSyncPointsAfter` is just making sure the the last BundleEndTime is greater than the last `EventStart`.

However, this does highlight an underlying assumption that the bundle time ranges are always day aligned. We'll need to be careful about this as we move to the new API. Maybe we just keep enforcing that the time ranges/durations are day aligned.

claygoddard: If we are always using `-1` though, do we run into problem when checkpoints are deleted that aren't day aligned? If there is only ever 1 entry, won't that trigger us to start back from the beginning rather than just start of day?

camphillips22: > If there is only ever 1 entry

I'm pretty sure that it's just inserting rows, not overwriting them, so there will be a row for each sync.

claygoddard: That's what I'm asking though. `-1` will give a unique ID automatically? Or is `ID` not even unique at all and all of these warehouses use some other autogenned identifier? I think that's the part I am missing.

camphillips22: `ID` isn't a special field, it's just an integer. And it doesn't look like bigquery or redshift treat them special.

claygoddard: Ah, okay, that was the part I was missing!

claygoddard (approved):
LG pending comment resolution.

PR #94, opened 2020-09-23
Allow headers with CSV format

### LINKED ISSUE #80: CSV headers not written when using LocalStorage option
When using local storage:

```
[local]
SaveDir = "/Users/van/Desktop/exports"
# Start time for data exports in the following format: 2018-12-27T18:30:00Z
StartTime = 2019-11-03T07:00:00Z
UseStartTime = false
```

The written CSV files lack headers (i.e. column names).  The quick fix appears to be simply adding a `Write` with the existing `tableColumns` in `ProcessFilesByDay` and `ProcessFilesIndividually`.

```
	csvOut := csv.NewWriter(outfile)

	// write the columns as CSV header
	csvOut.Write(tableColumns)
	csvOut.Flush()	// this flush may not be needed as flush later gets called
```

Is there any reason that we should not write headers in this scenario?  I haven't submitted a PR as I think we might make this configurable if headers should not be present in CSVs to work correctly with other adapters.

camphillips22:
Add headers to the CSV files. This allows you to use hauser to download CSV files locally
and make sense of them.

Fixes #80


claygoddard (approved):
LGTM

PR #97, opened 2020-12-09
Add multi-hauser recipe

camphillips22:
Adds a simple example for how to run multiple hauser processes in the same
docker container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,22 @@ FROM myexport
 WHERE JSON_EXTRACT_PATH_TEXT(CustomVars, 'acct_adminDisabled_bool') = 'false';
 ```
 
+## Dockerizing Hauser
+To add a specific `hauser` binary version to a Docker container (assuming that the base image is linux),
+simply add the following to the Dockerfile:
+
+```Dockerfile
+RUN curl -L >hauser.tar.gz https://github.com/fullstorydev/hauser/releases/download/v${HAUSER_VERSION}/hauser_${HAUSER_VERSION}_linux_x86_64.tar.gz \
+  && tar -xzvf hauser.tar.gz -C /usr/bin \
+  && rm hauser.tar.gz
+```
+Where `${HAUSER_VERSION}` is the desired release version.
+(This value can also be provided at build time with the [docker ARG command](https://docs.docker.com/engine/reference/builder/#arg).)
+
+For an example of a dockerized hauser solution that dynamically creates the build config files from a template,
+see [this recipe](./recipes/multi-hauser/README.md).
+
+
 ## Building from source
 * Make sure you have [installed](https://golang.org/doc/install) Go 1.11 or higher.
 * **OPTIONAL**: Set a custom [GOPATH](https://github.com/golang/go/wiki/SettingGOPATH).
--- a/README.md
+++ b/README.md
@@ -113,19 +113,18 @@ WHERE JSON_EXTRACT_PATH_TEXT(CustomVars, 'acct_adminDisabled_bool') = 'false';
 ```
 
 ## Dockerizing Hauser
-To add a specific `hauser` binary version to a Docker container (assuming that the base image is linux),
-simply add the following to the Dockerfile:
+To include Hauser in a Docker container, add the following to your Dockerfile.
+Note that the example below assumes the base image is linux.
 
 ```Dockerfile
 RUN curl -L >hauser.tar.gz https://github.com/fullstorydev/hauser/releases/download/v${HAUSER_VERSION}/hauser_${HAUSER_VERSION}_linux_x86_64.tar.gz \
   && tar -xzvf hauser.tar.gz -C /usr/bin \
   && rm hauser.tar.gz
 ```
-Where `${HAUSER_VERSION}` is the desired release version.
-(This value can also be provided at build time with the [docker ARG command](https://docs.docker.com/engine/reference/builder/#arg).)
 
-For an example of a dockerized hauser solution that dynamically creates the build config files from a template,
-see [this recipe](./recipes/multi-hauser/README.md).
+The `${HAUSER_VERSION}` can be provided at build time with the [docker ARG command](https://docs.docker.com/engine/reference/builder/#arg).
+You can find the latest version of Hauser on the [releases](https://github.com/fullstorydev/hauser/releases/latest) page.
+For a more complete example of using Hauser with docker, see [this recipe](./recipes/multi-hauser/README.md).
 
 
 ## Building from source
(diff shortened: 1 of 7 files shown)


recipes/multi-hauser/README.md
```diff
@@ -0,0 +1,72 @@
+# Multi-Hauser with Docker
+
+If your FullStory account has multiple associated accounts, this recipe can be used as a template to
+start a docker container that exports the data for each "org" and uploads it to BigQuery.
+While this is intended to work with Google Cloud, it can easily be adapted to work with
+AWS or any databases supported in the future.
+
+## How it works
+This recipe works by creating a docker container that uses [supervisord] to run multiple `hauser`
```
van-fs, line 9: Is the link to supervisord perhaps missing?  I see the brackets so I figure it was intended to be a link.

recipes/multi-hauser/README.md
```diff
@@ -0,0 +1,72 @@
...
+24 hours, 24 hours and 10 minutes, and 24 hours and 20 minutes, respectively.
+
+## Key File
+The key file should be a CSV with the following format:
+
+```
+<org1>,<api_key_for_org1>
+<org2>,<api_key_for_org2>
+<org3>,<api_key_for_org3>
+...
```
van-fs, line 39: I think the CSV file must have a new line at the end or it'll drop the last line.  (I had tried this approach with a single line with org:apikey pair, and it didn't work).  So perhaps a tip or nota bene related to this requirement could be added.

recipes/multi-hauser/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM alpine:latest
```
claygoddard, line 1: I'm pretty science dog on the docker stuff -- is there someone else that could look at this too? I mean I trust it works and its just a recipe, so not a huge deal either way.

van-fs: @claygoddard FWIW, I had written a similar dockerfile.  I like Cam's version more because it delegates construction of the artifacts to the container itself.  Previously, I had a build script that then pushed the output to the container.  FullStory customers have also created similar approaches https://github.com/fullstorydev/hauser/pull/12/files and https://github.com/aclaimant/docker-hauser/blob/master/Dockerfile.  @camphillips22, with this recipe, do you think that Schuylr's PR should be closed in reference to this?

claygoddard (approved):
I like the idea. I'll go ahead and approve, of course pending Van's comments!

recipes/multi-hauser/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN apk update && apk add --no-cache \
+  supervisor \
+  curl \
+  gettext
+
+ARG HAUSER_VERSION=1.0.0
```
van-fs, line 7: To future proof and support newer releases without updating the recipe, you could also get the latest download URL with this `HAUSER_DOWNLOAD=$(curl -s https://api.github.com/repos/fullstorydev/hauser/releases/latest | grep "browser_download_url.*linux_x86_64.tar.gz" | cut -d : -f 2,3 | tr -d \")`.  Though if we want to keep it simple, what we have is pretty clear.

recipes/multi-hauser/README.md
camphillips22: It's not missing (see the references at the bottom), but it is broken! Looks like I need to add `http://` to make sure it's not a relative link.

recipes/multi-hauser/README.md
camphillips22: Ooh, good call. My editors add that automatically so I didn't realize I needed to call it out.

recipes/multi-hauser/Dockerfile
camphillips22: I guess it could be. Honestly, I think that providing the binaries obviates the need for a dedicated docker image since it's easy enough to add the binary to the image or just to curl it like this. Especially since hauser depends on setting up Auth and using secrets and that can be quite different depending on the deployment environment. Maybe I could add a one-liner to the main README with a command to add the latest version to a Dockerfile.

recipes/multi-hauser/Dockerfile
camphillips22: Yeah, I saw that solution too, but went with this option since it was deterministic. If you feel strongly, I can update it   to be latest.

van-fs: Nah, just wanted to float it.  Seems like one more thing that will eventually break, and it's far more clear how to simply update the version value.

README.md
```diff
@@ -112,6 +112,22 @@ FROM myexport
...
+  && tar -xzvf hauser.tar.gz -C /usr/bin \
+  && rm hauser.tar.gz
+```
+Where `${HAUSER_VERSION}` is the desired release version.
+(This value can also be provided at build time with the [docker ARG command](https://docs.docker.com/engine/reference/builder/#arg).)
+
+For an example of a dockerized hauser solution that dynamically creates the build config files from a template,
+see [this recipe](./recipes/multi-hauser/README.md).
+
+
```
camphillips22, line 130: @claygoddard @van-fs Just added this per the comment about closing Schuylr's PR. Does this seem like a reasonable "solution", or do you think we need to have more guidance for a single hauser deployment?

van-fs: I find the above helpful.  I word-smithed it below to be a bit more prescriptive for comparison.  I feel like if we needed to be any more specific, we'd just defer to the multi-instance Hauser container.  The only difference between single and multi is the number of lines in the CSV.  (Obviously, there are technical implications, but from a narrative, it seems more understandable to have a single point of view versus two).  Long way of saying, short and sweet.

## Dockerizing Hauser
To include Hauser in a Docker container, add the following to your Dockerfile. Note that the example below assumes the base image is linux.

```Dockerfile
RUN curl -L >hauser.tar.gz https://github.com/fullstorydev/hauser/releases/download/v${HAUSER_VERSION}/hauser_${HAUSER_VERSION}_linux_x86_64.tar.gz \
  && tar -xzvf hauser.tar.gz -C /usr/bin \
  && rm hauser.tar.gz
```

The `${HAUSER_VERSION}` can be provided at build time with the [docker ARG command](https://docs.docker.com/engine/reference/builder/#arg).  You can find the latest version of Hauser on the [releases](https://github.com/fullstorydev/hauser/releases/latest) page.  For a more complete example of dockerizing Hauser, see [this recipe](./recipes/multi-hauser/README.md).

camphillips22: Ok, I updated it to use your word-smithing!

PR #98, opened 2020-12-14
Initial commit for Snowflake compatibility

jh-inst:
Hoping to get Snowflake compatibility implemented for my team.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1,107 +1,242 @@
-package client
+package config
 
 import (
-	"bytes"
+	"errors"
 	"fmt"
-	"io"
-	"net/http"
-	"strconv"
+	"io/ioutil"
+	"log"
 	"time"
 
-	"github.com/fullstorydev/hauser/config"
+	"github.com/BurntSushi/toml"
 )
 
-var _ error = StatusError{}
+const (
+	// DefaultApiURL is the standard base URL for the fullstory.com API.
+	DefaultApiURL         = "https://api.fullstory.com"
+	DefaultExportDelay    = 24 * time.Hour
+	DefaultExportDuration = 1 * time.Hour
+	MinExportDuration     = 15 * time.Minute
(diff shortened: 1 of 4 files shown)


van-fs:
Thanks @jh-inst .  We'll take a look, and definitely the right time of the year for this PR ❄️ ❄️ .

client/client.go
```diff
@@ -1,107 +1,242 @@
-package client
+package config
```
camphillips22, line 1: It looks like these changes were an unintentional copy of the `config.go` file?

config/config.go
```diff
@@ -91,6 +93,19 @@ type RedshiftConfig struct {
...
+	DB             string
+	User           string
+	Password       string
+	ExportTable    string
+	SyncTable      string
+	DatabaseSchema string
+	Credentials    string
+	VarCharMax     int
+	S3Region       string `toml:"-"`
+}
```
camphillips22, line 107: This has the exact same structure as `Redshift`. I haven't worked with snowflake, but if this is all that's necessary, we should just merge these concepts into a single thing. Technically, the fact that it's labelled `Redshift` at all is misleading since it just assumes some database that supports the postgres driver.

Also, looking at `snowflake.go`, it's the same code as `redshift.go`. I'm curious, if you just plug in your values that you're using for snowflake the redshift config, does it just work? This would be great since it wouldn't require any changes to the source (aside from documentation, and possibly naming for clarity).

camphillips22 (commented):
Thank you @jh-inst for your PR. I left some comments, but high level it looks like all of these changes are unnecessary (assuming that a snowflake DB can just be loaded with the `postgres` SQL driver).

van-fs:
@jh-inst , based on our chat, it sounded like this approach may need to be revisited.  Should we close this PR and work from a new one?

jh-inst:
@van-fs - yes, go ahead and close.  Thank you!

PR #99, opened 2021-02-15
PageVars: Add to hauser!

theMPatel:
We added new custom scoped variables, dubbed page vars. This PR adds functionality to include those if present so they show up in the export


camphillips22:
Can you update the test data and verify that the extra variables are formatted as desired?

theMPatel:
PTAL

camphillips22 (approved):
Sorry for taking so long to get back around to this! LGTM!

PR #100, opened 2021-02-15
Replace travis CI with CircleCI

camphillips22:
We ran out of credits on our OSS plan, so we'll just use CircleCI.


.circleci/config.yml
```diff
@@ -0,0 +1,25 @@
+# Golang CircleCI 2.0 configuration file
+#
+# Check https://circleci.com/docs/2.0/language-go/ for more details
+version: 2
+jobs:
+  build:
+    docker:
+      # specify the version
+      - image: circleci/golang:1.14
```
camphillips22, line 9: @claygoddard Do you think we need to add any more versions? I was thinking that since we release a binary, testing multiple versions isn't that important.

claygoddard: This is fine with me!

claygoddard (approved):
Nice!

PR #101, opened 2021-03-11
Add FilePrefix to configuration

camphillips22:
This enables a "prefix" to be specified in the config file.
This configuration is especially useful if a user does not want
the files that hauser creates to be uploaded to the "root" folder
in cloud storage.

### DIFF
--- a/example-config.toml
+++ b/example-config.toml
@@ -31,6 +31,12 @@ Provider="local"
 StorageOnly = false
 SaveAsJson = false
 
+# FilePrefix can be used to specify a prefix for each of the files that are created.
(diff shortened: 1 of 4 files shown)


example-config.toml
```diff
@@ -31,6 +31,12 @@ Provider="local"
 StorageOnly = false
 SaveAsJson = false
 
+# FilePrefix can be used to specify a prefix for each of the files that are created.
+# For example, if using GCS or S3, you can use this setting to organizer your hauser uploads
```
claygoddard, line 35:
```suggestion
# For example, if using GCS or S3, you can use this setting to organize your hauser uploads
```

claygoddard (approved):
Nice!

PR #102, opened 2021-04-01
Fix FilePrefix when StorageOnly=true

camphillips22:
With the addition of #101, I missed that the
sync file that is created when `StorageOnly=true`
did not also receive the `FilePrefix` setting.


claygoddard (approved):
LGTM

PR #103, opened 2021-04-27
Add request fields

camphillips22:
Exports now include the following fields when the `EventType`
is `request`:

* `ReqUrl`: the URL for the corresponding XHR request
* `ReqMethod`: the HTTP request method for the corresponding XHR request
* `ReqStatus`: the HTTP response status code for the corresponding XHR request


claygoddard (approved):
LGTM!

PR #105, opened 2021-05-20
Add code scanning workflow

camphillips22:
Let's level up the security for this project by adding code scanning.


camphillips22:
In the checks, the

> 1 analysis not found

bit just means that the base branch doesn't have an analysis so it can't do a comparison.

claygoddard (approved):
LGTM!

PR #106, opened 2021-05-21
Dockerized hauser

camphillips22:
Add a dockerfile that can be used to build
a minimal docker image. Also adds a script for
releasing based on the pattern established in
[grpcurl](https://github.com/fullstorydev/grpcurl) that
makes the release and pushes a docker image.


jhump (approved):
LGTM!

PR #107, opened 2021-05-28
Fix folder permissions in docker

camphillips22:
There was an issue with using the scratch
image because hauser expects to be able to create
directories for temporary files. With the scratch
image, the `hauser` user did not have permission to
create the directory, so the process would fail.
Use `--chown` flag when copying over the directory
to ensure the `hauser` user can write to it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,9 @@
 FROM golang:1.13-alpine as builder
 MAINTAINER FullStory Engineering
 
-# create non-priveleged group and user
-RUN addgroup -S hauser && adduser -S hauser -G hauser
+# create non-priveleged group and user and an owned directory
+RUN addgroup -S hauser && adduser -S hauser -G hauser && \
+ mkdir /hauser && chown -R hauser:hauser /hauser
 
 WORKDIR /tmp/fullstorydev/hauser
 COPY VERSION *.go go.* /tmp/fullstorydev/hauser/
@@ -16,16 +17,17 @@ ENV CGO_ENABLED=0
 ENV GOOS=linux
 ENV GOARCH=amd64
 ENV GO111MODULE=on
-RUN go build -o /hauser \
+RUN go build -o /bin/hauser \
     -ldflags "-w -extldflags \"-static\" -X \"main.version=$(cat VERSION)\"" \
     .
 
-FROM alpine:3.13.5
-WORKDIR /hauser
-RUN addgroup -S hauser && adduser -S hauser -G hauser && chown -R hauser:hauser /hauser
+FROM scratch
 COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 COPY --from=builder /etc/passwd /etc/passwd
-COPY --from=builder /hauser /bin/hauser
+COPY --from=builder /etc/group /etc/group
+COPY --from=builder /bin/hauser /bin/hauser
+COPY --from=builder --chown=hauser /hauser /hauser
+WORKDIR /hauser
 USER hauser
 
 ENTRYPOINT ["/bin/hauser"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,8 +20,9 @@ RUN go build -o /hauser \
     -ldflags "-w -extldflags \"-static\" -X \"main.version=$(cat VERSION)\"" \
     .
 
-FROM scratch
-WORKDIR /
+FROM alpine:3.13.5
+WORKDIR /hauser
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -20,8 +20,9 @@ RUN go build -o /hauser \
     -ldflags "-w -extldflags \"-static\" -X \"main.version=$(cat VERSION)\"" \
     .
 
-FROM scratch
-WORKDIR /
+FROM alpine:3.13.5
+WORKDIR /hauser
+RUN addgroup -S hauser && adduser -S hauser -G hauser && chown -R hauser:hauser /hauser
```
jhump, line 25: Could you do this in the builder image instead and just copy over the directory with owner/perms in tact? Then you could continue to use scratch and have a much smaller final image.

jhump: I see you already copy the `/etc/passwd` file over from the builder below. I think you might also want to copy over `/etc/group`.

camphillips22: So when I first tried this, it didn't work, but it looks like I was missing the `--chown` flag for COPY. I'll update! And good point on adding `/etc/group`!

camphillips22:
@jhump PTAL

Dockerfile
```diff
@@ -1,8 +1,9 @@
 FROM golang:1.13-alpine as builder
 MAINTAINER FullStory Engineering
 
-# create non-priveleged group and user
-RUN addgroup -S hauser && adduser -S hauser -G hauser
+# create non-priveleged group and user and an owned directory
```
jhump, line 4:
```suggestion
# create non-privileged group and user and an owned directory
```

jhump: Maybe worth a comment that we do this in the builder, instead of below in the target container, because "scratch" doesn't include these utilities.

Dockerfile
```diff
@@ -16,15 +17,17 @@ ENV CGO_ENABLED=0
...
 
 FROM scratch
-WORKDIR /
 COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 COPY --from=builder /etc/passwd /etc/passwd
-COPY --from=builder /hauser /bin/hauser
+COPY --from=builder /etc/group /etc/group
+COPY --from=builder /bin/hauser /bin/hauser
+COPY --from=builder --chown=hauser /hauser /hauser
+WORKDIR /hauser
```
jhump, line 30: Perhaps a README update is warranted, to indicate that the config file used should omit the `TmpDir` param or, if present, it must be relative or start with `/hauser/`?

Is there any value in a container _preserving_ these files from run to run? Does hauser need to save a "cursor" to disk, so it can correctly resume after a restart? If so, then doing so just in the temporary volume won't be good since it means it would try to re-download all history if the container is restarted. (Maybe that cursor is saved somewhere else, like BQ?)

If it does need to be saved to the local file-system, then we'd also need directions about mounting that as a volume to the container, so it gets persisted in the host's file system. And to do that with a non-root user, we may want to pick a hard-coded UID and GID for the user/group. That way, users can do `chown :1024 ~/hauser-cursor` to make sure the container can write to their host file system (where "1024" in this example is the UID of the "hauser" user inside the container). (For more context: https://medium.com/@nielssj/docker-volumes-and-file-system-permissions-772c1aee23ca).

camphillips22: > Perhaps a README update is warranted, to indicate that the config file used should omit the TmpDir param or, if present, it must be relative or start with /hauser/?

I think this is necessary regardless. I can make an update with this PR to include some guidance for running the hauser image.

> Is there any value in a container preserving these files from run to run? Does hauser need to save a "cursor" to disk, so it can correctly resume after a restart? If so, then doing so just in the temporary volume won't be good since it means it would try to re-download all history if the container is restarted. (Maybe that cursor is saved somewhere else, like BQ?)

In short, no. If the process is running in "local" mode, then it will save the cursor to wherever the files are being stored, which should be in a mounted folder anyways. For all other processes, the cursor is stored in the destination (a dedicated table for database destinations and a cursor file for storage destinations).

example-config.toml
```diff
@@ -1,7 +1,11 @@
 FsApiToken = "<your FullStory API token>"
 Backoff = "30s"
 BackoffStepsMax = 8
-TmpDir = "/tmp"
+
+# TmpDir is a directory where the exported files are downloaded to before they are uploaded
+# to the warehouse destination. The hauser process will remove these files when it has finished
+# processing it.
```
jhump, line 7: supernit
```suggestion
# processing them.
```

jhump (approved):
LGTM!

PR #108, opened 2021-05-28
Add GKE recipe

camphillips22:
This recipe provides a script that deploys a recommended
hauser deployment to GKE, including a service account
with "least privilege".

### DIFF
--- a/recipes/GKE Setup/README.md
+++ b/recipes/GKE Setup/README.md
@@ -1,0 +1,105 @@
+# Google Kubernetes Engine (GKE) Setup
+
+There are many ways to set up hauser to run in a cloud environment; this guide does not
+aim to be comprehensive, but rather to provide a recommended configuration for a simple workflow.
+
+In order to be able to set up the infrastructure for your project, it's best if you have at an [Editor
+role](https://cloud.google.com/iam/docs/understanding-roles#basic). Else, you will need permission to
+create/edit the following resources:
+* Google Cloud Storage (GCS) buckets
+* BigQuery Datasets
+* GKE Kubernetes objects
+* Service Accounts
+* Service Account IAM Policies
+
+Ensure that you have the [Google Cloud SDK](https://cloud.google.com/sdk/docs/install) and `kubectl` installed
+and that you've [enabled cluster access](https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-access-for-kubectl).
+
+To run the `./deploy.sh` script, you must also have the following CLI tools installed and in your $PATH
+* `gsutil` - Storage CLI (installable with `gcloud components install gsutil`)
+* `bq` - BigQuery CLI (installable with `gcloud components install bq`)
+* [`jq`](https://stedolan.github.io/jq/) - JSON processor
+
+## Prerequisites
+As described in the [cluster access documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-access-for-kubectl),
+first, ensure that your gcloud SDK is initialized:
+```shell
+gcloud init
+```
+This will authenticate the SDK, allow you to select a project, and configure a default zone/region.
+
+Then (assuming you are using an existing cluster) you must generate a kubernetes configuration file:
+```shell
+gcloud container clusters get-credentials "<cluster-id>"
+```
+Before running the `deploy.sh` script, you must
+[create a BigQuery dataset](https://cloud.google.com/bigquery/docs/datasets#create-dataset) (or use an existing dataset)
+and a [Google Storage bucket](https://cloud.google.com/storage/docs/creating-buckets).
+
+As a default, the script expects the name of the dataset to be `fullstory-export-data` and the
+name of the bucket to be `$GOOGLE_PROJECT-hauser`, but feel free to change these.
+
+If you did not use the default dataset and bucket names, or want to customize the name of the service account,
+feel free to edit the following variables at the top of the `deploy.sh` script.
+```shell
+BIGQUERY_DATASET="fullstory-export-data"
+SERVICE_NAME="hauser-svc"
+GCS_BUCKET="$GOOGLE_PROJECT-hauser"
+HAUSER_VERSION="latest"
+```
+To find your FullStory Org Id, see [help docs](https://help.fullstory.com/hc/en-us/articles/360047075853-How-do-I-find-my-FullStory-Org-Id-).
+You will also need to create or get a [FullStory API Key](https://help.fullstory.com/hc/en-us/articles/360020624834-Where-can-I-find-my-API-key-).
+
+The `HAUSER_VERSION` variable corresponds to the docker image tag. See [docker hub](https://hub.docker.com/r/fullstorydev/hauser) for available tags.
+
+## The deploy script
+The `deploy.sh` script in this directory uses the gcloud SDK and associated CLI's to
+create a service account, add the permissions, and create a Kubernetes deployment that
+runs hauser.
+
+Assuming you have completed the prerequisites, you can run the deploy script with:
+```shell
+./deploy.sh -p "<my_project_id>" -o "<my_org_id>" -k "<fs_api_key"
+```
+
+Note: The script is idempotent, so if it fails, you can simply run it again after resolving any issues.
+
+If a service account does not exist, the script will confirm before creating it. Assuming the service account creation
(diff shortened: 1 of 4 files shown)


recipes/GKE Setup/README.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+To verify that everything is working as expected, you can also check the logs with
+```shell
+kubect logs -lapp=hauser -c hauser
+```
+
+## Service accounts
+This setup is based on using a service account for the hauser deployment.
+In short, this is recommended to adhere to the principal of east privilege.
```
jhump, line 88:
```suggestion
In short, this is recommended to adhere to the principal of least privilege.
```

recipes/GKE Setup/README.md
```diff
@@ -0,0 +1,105 @@
...
+```shell
+BIGQUERY_DATASET="fullstory-export-data"
+SERVICE_NAME="hauser-svc"
+GCS_BUCKET="$GOOGLE_PROJECT-hauser"
+HAUSER_VERSION="latest"
+```
+To find your FullStory Org Id, see [help docs](https://help.fullstory.com/hc/en-us/articles/360047075853-How-do-I-find-my-FullStory-Org-Id-).
+You will also need to create or get a [FullStory API Key](https://help.fullstory.com/hc/en-us/articles/360020624834-Where-can-I-find-my-API-key-).
+
+The `HAUSER_VERSION` variable corresponds to the docker image tag. See [docker hub](https://hub.docker.com/r/fullstorydev/hauser) for available tags.
```
jhump, line 53: Maybe mention that the latest available version will be used by default?

jhump: Also, might move this up above to line 40-ish. That way you introduce all of the parameters and their defaults first, and then the following paragraph describes how to change them by editing values in the script.

recipes/GKE Setup/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+## The deploy script
+The `deploy.sh` script in this directory uses the gcloud SDK and associated CLI's to
+create a service account, add the permissions, and create a Kubernetes deployment that
+runs hauser.
+
+Assuming you have completed the prerequisites, you can run the deploy script with:
+```shell
+./deploy.sh -p "<my_project_id>" -o "<my_org_id>" -k "<fs_api_key"
```
jhump, line 62: It's usually bad practice to stick secrets in command-line args. Perhaps this should accept the API key via env var instead, and then complain at script startup if the env var is unset?

(Also, nit, missing a trailing `>`.)

recipes/GKE Setup/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Service accounts
+This setup is based on using a service account for the hauser deployment.
+In short, this is recommended to adhere to the principal of east privilege.
+To read more about service accounts in Google, see [Google's documentation](https://cloud.google.com/iam/docs/service-accounts)
+
+The `create-svc-account.sh` script will create/validate that the service account
+exists and has all of the necessary permissions and IAM policies associated with it.
+This includes:
+* [Workload Identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity) to associate the gcloud
```
jhump, line 94: Do we need content about how to configure the workload if they are _not_ using workload identity?

That is a bit more involved, unfortunately, requiring one to provision a JSON key, upload it as a k8s secret and then edit the deployment to mount the secret and then point to the file therein via `GOOGLE_APPLICATION_CREDENTIALS` env var.

recipes/GKE Setup/create-svc-account.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+
+set -e
+
+SERVICE_EMAIL="$SERVICE_NAME@$GOOGLE_PROJECT.iam.gserviceaccount.com"
```
jhump, line 5: The README should probably mention that `SERVICE_NAME` must be set before invoking. Or perhaps this should accept it as an argument?

Also, this script should definitely validate that it's not empty before proceeding.

jhump: `GOOGLE_PROJECT`, `BIGQUERY_DATASET`, and `GCS_BUCKET`, too

recipes/GKE Setup/README.md
```diff
@@ -0,0 +1,105 @@
...
+```shell
+kubect logs -lapp=hauser -c hauser
+```
+
+## Service accounts
+This setup is based on using a service account for the hauser deployment.
+In short, this is recommended to adhere to the principal of east privilege.
+To read more about service accounts in Google, see [Google's documentation](https://cloud.google.com/iam/docs/service-accounts)
+
+The `create-svc-account.sh` script will create/validate that the service account
```
jhump, line 91: It doesn't look like users are intended to run this directly, but instead just let `deploy.sh` run it. Is that right? Maybe merge into `deploy.sh` if that is the case? Or document it a little more (required inputs/params) and/or make it more usable to be run independently?

recipes/GKE Setup/deploy.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/bash
+
+set -e
+
+GOOGLE_PROJECT=""
+ORG_ID=""
+BIGQUERY_DATASET="fullstory-export-data"
+SERVICE_NAME="hauser-svc"
+GCS_BUCKET="$GOOGLE_PROJECT-hauser"
```
jhump, line 9: I think the script needs to be tested in an empty/dummy project. For example,`GOOGLE_PROJECT` var is empty at this point, which means this script cannot work.

jhump: You could remove the command-line args and make the interface entirely driven by env vars. Then you can rely on `GOOGLE_PROJECT` being set (since the script will abort if it is not). To preserve defaults for the ones below, you can do this:
```sh
BIGQUERY_DATASET="${BIGQUERY_DATASET:-fullstory-export-data}"
```

recipes/GKE Setup/hauser.yaml
```diff
@@ -0,0 +1,72 @@
+---
```
jhump, line 1: I wonder if this would be better to provide as a helm chart and vars file and a terraform module...

@nrmitchi and @bschaeffer may have thoughts.

recipes/GKE Setup/README.md
camphillips22: Yeah, I was vacillating on that. It actually already works if you don't provide it. I'll just remove it from the args and require that it's set before running the script.

recipes/GKE Setup/README.md
camphillips22: Yeah, this is probably the part I'm least clear about. Just so I understand, adding the service account annotation will not work if WID is not enabled? I was going to do the whole "create a json key and add it to the secret" thing at first, but went this route first since it's a bit simpler. Maybe we prompt to either use WID or use private key directly?

recipes/GKE Setup/README.md
camphillips22: Yeah, I think I might just merge it. That'll definitely simplify the validation bits.

recipes/GKE Setup/hauser.yaml
nrmitchi: I would agree. A Helm chart here would typically accept a service account to use as a parameter, and the GCP service account would be linked elsewhere. This would allow the basic kubernetes definitions/deployment to be applicable in any kubernetes cluster, and not specific to GKE.

camphillips22: I definitely think there's value in adding these, but I wanted to start with a "minimal set" of deps (just gcloud and jq) so that I could get a working GKE recipe quicker. Also, I'm not as familiar with setting up the helm chart (typically in it's own repo?) and I haven't worked with the other clouds enough to know how to write that more generally.

camphillips22: But maybe it's simpler than I'm imagining. Just haven't done the research here yet.

nrmitchi: I don't think a Helm chart that encompasses what you have here would be difficult; it would mostly be wrapping the manifests you've already written, and updating the documentation.

I don't think the lack-of-helm should be considered a blocker here, but we should come back and add it. This is because hauser is safe to remove and recreate, which would make a switch to helm easier for anyone using this.

PR #109, opened 2021-06-02
Add options to modify client

camphillips22:
When used as a library, it can be useful to modify
how the client operates. This change adds the ability
to modify the client using the "options" pattern.
The two new options are:
* WithHttpClient - for modifying the underlying http transport
* WithCreateExportRequestModifier - for modifying the "CreateExport"
  API request before sending.


claygoddard (approved):
LGTM!

PR #110, opened 2021-06-03
Allow SaveAsJson=true when StorageOnly=true

camphillips22:
There is currently a requirement that `SaveAsJson`
is only available if the provider is local. This
should be available for all providers assuming
`StorageOnly=true`.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -209,10 +209,6 @@ func Validate(conf *Config, getNow func() time.Time) error {
 		conf.Warehouse = ""
 	}
 
-	if conf.SaveAsJson && conf.Provider != "local" {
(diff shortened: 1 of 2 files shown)


claygoddard (approved):
LGTM

drew88:
is there any plans to release this change?  we are currently running v1.2.0 but would like to use JSON for StorageOnly=true

camphillips22:
@drew88 We don't have an official cadence for releasing, but I can go ahead and make a new one.

camphillips22:
@drew88 Late on a Friday, but I've created a new release for version [1.3.0](https://github.com/fullstorydev/hauser/releases/tag/v1.3.0). Happy hacking!

PR #112, opened 2021-10-06
Update README.md and example-config.toml to be more accurate

randito78:
Added a note about necessary FullStory API permission levels for the API key as well as change the note for StartTime as it errors with the pre-existing empty string while casting to time with the error `Error: cannot parse "'2020-09-01T20:41:15Z'" as "".`

Recently went through setting up hauser for my business and ran into a couple of issues to I wanted to make sure no one else did as well.

### DIFF
--- a/example-config.toml
+++ b/example-config.toml
@@ -22,9 +22,7 @@ ExportDuration = "6h"
 ExportDelay = "24h"
 
 # StartTime determines how far back to start exporting data if starting fresh.
-# This should be an timestamp like with the followin format: 2018-12-27T18:30:00Z.
-# If start time is empty, this will default to 30 days in the past.
-StartTime = ""
+StartTime = <Start time for data exports in the following format: 2018-12-27T18:30:00Z>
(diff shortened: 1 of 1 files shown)


example-config.toml
```diff
@@ -22,9 +22,7 @@ ExportDuration = "6h"
 ExportDelay = "24h"
 
 # StartTime determines how far back to start exporting data if starting fresh.
-# This should be an timestamp like with the followin format: 2018-12-27T18:30:00Z.
-# If start time is empty, this will default to 30 days in the past.
-StartTime = ""
+StartTime = <Start time for data exports in the following format: 2018-12-27T18:30:00Z>
```
camphillips22, line 25: Instead of using a value that is definitely not parseable, I would prefer that we use the "zero" time value for go

```suggestion
# If not set or left as the zero time, this will default to 30 days in the past.
# StartTime = "0001-01-01T00:00:00Z"
```

PR #114, opened 2022-02-10
Build improvements

camphillips22:
Cleans up non-module artifacts:
* Deletes vendor folder
* Fixes makefile for CLI installation
* go mod tidy


Makefile
```diff
@@ -1,11 +1,4 @@
 dev_build_version=$(shell git describe --tags --always --dirty)
-# to_check is all code in this repo that we want to run checks on
-# (it is all Go code in here, but intentionally excludes the
-# vendor folder contents)
-dirs_to_check=$(shell find . -maxdepth 1 -mindepth 1 -type d | grep -vE '\./\.|vendor|dist|recipes|releasing')
```
jameremo, line 5: There's more than just the vendor directory in the original exclude list; were you thinking that the simplicity of the Makefile was more important than excluding them, now that most of the excluded files are gone? Or was there a specific problem this was solving?

jameremo (approved):
Overall LGTM, I had one question just to make sure I understand some of the Makefile simplifications. It's kind of cool to see us close the interim chapter on go modules here!

Makefile
camphillips22: It's more that this was frustrating if we added any more folders because I would then have to change this. For example, I ran into issues with testing because I had a local `tmp` directory and this formulation meant that the CLIs were run for each folder that was found in the base directory. So in the test case, this was explicitly calling

```bash
go test -race . ./core ./config ./internal ./testing ./warehouse ./tmp ./client
```

which would fail with `no Go files in /Users/cam/git/hauser/tmp`.

`go test -race ./...` works just fine with folders that have no go files in them and it also includes any new folder without having to update this nastiness.

camphillips22: Now, we couldn't run `go test -race ./...` with the vendored directory since it would check all of the vendored files. So removing the vendor dir meant that we didn't need this check at all.

PR #124, opened 2023-12-04
Add Added SegmentId to example-config.toml

rcmaples:
Added SegmentId to the example toml. @camphillips22 Can you double check I added the right syntax? Is there any other comment needed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,13 @@ At this point, `hauser` will perform the process approximately every `ExportDura
 When using a database, it uses the `SyncTable` to keep track of what export files have been processed, and will restart from the last known sync point.
 For a `StorageOnly` process, it will create a file called `.sync.hauser` that will be used as a checkpoint.
 
+### Syncing Specific Segments
+_Note: this field is optional and should only be set if syncing a specific segment._
+
+By default `hauser` will sync the "Everyone" segment. If you would rather sync a specific segment, you can specify a value by setting the `SegmentId` config option. You can use the [list segment endpoint](https://developer.fullstory.com/server/v1/segments/list-segments/) to get a list segment endpoint to get a list of IDs of segments in your org.
+
+Alternatively, the segment's URL in FullStory will contain the segment's ID. For example, a segment with the following URL `https://app.fullstory.com/ui/ABCD/segments/alice@example.com:12444435241/people/0` has a segment ID of `alice@example.com:12444435241`. You would then add `SegmentId = "alice@example.com:12444435241"` to your toml.
+
 ### Amazon Web Services Notes
 _Currently, only S3 and Redshift are supported for this provider._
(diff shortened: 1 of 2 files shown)


example-config.toml
```diff
@@ -2,6 +2,9 @@ FsApiToken = "<your FullStory API token>"
 Backoff = "30s"
 BackoffStepsMax = 8
 
+# SegmentId is the string identifier for the segment you want to use with Hauser. 
+SegmentId = "<your segment id>"
```
camphillips22, line 6: This value is not required and this does not result in a valid config, so I would comment this line out.

As far as a comment, is there a help doc somewhere referencing where a segment ID can be retrieved in the UI? Maybe point to that, or add a section in the README about how to determine this? I don't think it's necessary, but could help with discovery.

rcmaples: I revised the toml changes and added a section to the README. Let me know if it needs additional tweaking.

README.md
```diff
@@ -66,6 +66,13 @@ At this point, `hauser` will perform the process approximately every `ExportDura
 When using a database, it uses the `SyncTable` to keep track of what export files have been processed, and will restart from the last known sync point.
 For a `StorageOnly` process, it will create a file called `.sync.hauser` that will be used as a checkpoint.
 
+### Syncing Specific Segments
+_Note: this field is optional and should only be set if syncing a specific segment._
+
+By default `hauser` will sync the "Everyone" segment. If you would rather sync a specific segment, you can specify a value by setting the `SegmentId` config option. You can use the [list segment endpoint](https://developer.fullstory.com/server/v1/segments/list-segments/) to get a list segment endpoint to get a list of IDs of segments in your org.
+
+Alternatively, the segment's URL in FullStory will contain the segment's ID. For example, a segment with the following URL `https://app.fullstory.com/ui/ABCD/segments/alice@example.com:12444435241/people/0` has a segment ID of `alice@example.com:12444435241`. You would then add `SegmentId = "alice@example.com:12444435241"` to your toml.
```
camphillips22, line 74: Oof, I know that the dev docs use the email prefix, but this is the old version of our segment Ids. This has since been updated. Maybe change this to be something more random or state that older segment IDs may look like `X`?

rcmaples: Sorry Cam. I used this page: https://developer.fullstory.com/server/v1/segments/list-segments/ for the examples. Didn't realize it was v1. It's also hard to find the segment export stuff now on the dev docs. Looks like anything not v2 isn't shown in the nav bar (screenshot). 

Is the segment id highlighted in [this kb doc](https://help.fullstory.com/hc/en-us/articles/360057606414-HTTP-API-Using-Segment-Export#01ER5MMG4E8FZNTNB98XPD49JJ) correct? 

image

